PR #16, opened 2017-01-15
Add note for setting the APP_URI setting in .env

bobbybouwmann:
Hi,

When running Dusk the first time by just following the installation guide it will fail, since the APP_URI is not set correctly. Some of us use Hometead or Valet, so `http://localhost` is not valid. 

I added a small note, maybe you want to update the wording or the position, but I think we should at least point this out.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -39,7 +39,7 @@ To run your tests, use the `dusk` command. The `dusk` command accepts any argume
 
     php artisan dusk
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
It's not APP_URI. It's APP_URL.

bobbybouwmann:
Whoops, updated!

PR #24, opened 2017-01-16
Retain consistency of test methods with integrated

drawmyattention:
With Laravel's existing testing capabilities, the method signature is always ```$value```, ```$field```. The new dusk methods reverse this.

To make it easier to adopt, and apply convention I've standardised this.

The ```value()``` method is an exception to this however. I can resolve that using the following (which I can submit separately if you like):

```
/**
     * Directly get or set the value attribute of an input field.
     *
     * @param  string  $value
     * @param  string|null  $selector
     * @return $this
     */
    public function value($value, $selector = null)
    {
        if (func_num_args() == 1) {
            return $this->resolver->findOrFail($value)->getAttribute('value');
        }
        
        $selector = $this->resolver->format($selector);

        $this->driver->executeScript(
        );

        return $this;
    }
```

I'm not 100% happy with this however, as when one argument is passed, the parameter named ```$value``` is being used as a selector which makes readability a little off. I'd be happy to see alternative ways in how this could be tackled.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -56,7 +56,7 @@ $this->browse(function ($first, $second) {
     $second->loginAs(User::find(2))
             ->visit('/home')
             ->waitForText('Message')
-            ->type('message', 'Hey Taylor')
+            ->type('Hey Taylor', 'message')
             ->press('Send');
 
     $first->waitForText('Hey Taylor')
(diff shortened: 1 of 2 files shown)


taylorotwell:
Prefer selector first.

drawmyattention:
I think it might be good to get a wider consensus before rejecting this one. 

What the current signature means is we'll need to keep referring to the docs for _both_ the integrated testing, and dusk. Anyone new to the language would need to keep checking to figure out the order, any veterans to the framework would need to keep checking dusk docs rather than being able to apply their existing knowledge.

Changing the existing testing functionality would break b/c so really this would be the only solution.

PR #28, opened 2017-01-17
Remove jQuery dependency from seeLink

### LINKED ISSUE #27: jQuery requirement
Hello,

The following line: 
	
	var link = $("{$selector}");

in `Laravel\Dusk\Concerns\MakesAssertions`'s `seeLink()` function is problematic for projects that do not use jQuery as developers will be presented with `Facebook\WebDriver\Exception\UnknownServerException: unknown error: $ is not defined`.

Would it be possible to remove the jQuery dependency with native javascript?

fetzi:
This removes the jQuery dependency in the `seeLink` Method and uses Vanilla Javascript for checking if the link is displayed.

This PR resolves #27.

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -221,7 +221,7 @@ trait MakesAssertions
             return Array.prototype.slice
             .call(document.querySelectorAll("{$selector}"))
             .some(function(element) {
-              return element.offsetHeight > 0 && element.offsetWidth > 0 && element.style.visibility !== "hidden" ;
+              return element.offsetHeight > 0 && element.offsetWidth > 0;
             });
 JS;
 
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -215,15 +215,15 @@ trait MakesAssertions
      */
     public function seeLink($link)
     {
-      $script = <<<JS
-          return Array.prototype.slice
-          .call(document.querySelectorAll("a"))
-          .some(function(element) {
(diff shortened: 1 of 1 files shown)


src/Concerns/MakesAssertions.php
```diff
@@ -215,11 +215,14 @@ public function assertDontSeeLink($link)
      */
     public function seeLink($link)
     {
-        $selector = trim($this->resolver->format("a:contains('{$link}')"));
+        $selector = trim($this->resolver->format("a[href*='{$link}']"));
```
spyric: I think it's wrong replacement. 

for example  if I call
```
$browser->seeLink('Projects')
```
`a:contains('{$link}')` Will return link with content 'Projects'

But in your implementation it is looking in `href` attribute and not in text .

fetzi: that is true, one solution would be to check the innerHTML of the element. But I think another solution would be to split this one into two seperate methods `seeLinkText` and `seeLinkHref` @taylorotwell what do you think?

fetzi:
@spyric I've updated the `seeLink` method to match against the link text

taylorotwell:
This doesn't work in my simple test.

fetzi:
@taylorotwell can you please give me info what is not working? the implementation does the same as jquery does.

taylorotwell:
It just sat on the page and didn't do anything. I've pushed changes to just load in jQuery if it's not available.

fetzi:
is loading a library the right solution if it can easily be done with vanilla js? I will have a look on the implementation and send another PR again, if you would like to?

PR #33, opened 2017-01-18
Display Environmental variable fix for majority of Linux versions

trideout:
This fix will allow for most Linux versions to properly access the display to launch the chrome window.

### DIFF
--- a/src/SupportsChrome.php
+++ b/src/SupportsChrome.php
@@ -21,7 +21,7 @@ trait SupportsChrome
         if (PHP_OS === 'Darwin') {
             static::$chromeProcess = new Process('./chromedriver-mac', realpath(__DIR__.'/../bin'), null, null, null);
         } else {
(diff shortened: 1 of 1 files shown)


simonhunt:
Works perfectly for me (running Linux Mint and using Homestead).

trideout:
@simonhunt can you see if the change breaks support for you on Mint? I'm on Ubuntu and the console is not passing the environmental variables along so the error is `[25205:25205:0117/161540:ERROR:browser_main_loop.cc(265)] Gtk: cannot open display: `

simonhunt:
Sorry, I wasn't clear enough - the patch works perfectly! It didn't work before.

simonhunt:
Although, somewhat bizarrely the tests occasionally take 30 seconds to run (as opposed to < 1 second) - about 1 / 3 times it does this.

trideout:
@simonhunt excelent, thanks :)

aguirreg:
Where is the SupportsChrome.php file?

PR #34, opened 2017-01-18
Added lightweight css selector library

lkmadushan:
- Sizzle is a lightweight css selector engine provided by jQuery community and a component of jQuery
- [sizzlejs](https://sizzlejs.com/)


KKSzymanowski:
Why is using jQuery an issue for testing? It doesn't affect your application whatsoever and its loading time is nowhere near the booting time of phpunit and chromedriver. It's reliable, everyone knows how to use it, and everyone can easily debug it.

One issue which has to be addressed is possibility of conflict with other JS libraries registered under the dollar sign. This however has been already resolved with using `window.jQuery` instead of `window.$`.

Introducing a new, even 10x faster may cause more trouble than good, in my opinion.

lkmadushan:
@KKSzymanowski Sizzle is a component of jQuery and also jQuery includes lot of unwanted things that don't need for testing

fetzi:
If using a library is the way to go, I would also stick to jQuery. Otherwise I think rewriting to Vanilla JS would be a good solution.

KKSzymanowski:
>Sizzle is a component of jQuery

I did not know that. I rest my case.

taylorotwell:
What is the measurable difference in performance of using this vs just using jQuery. Please post results.

lkmadushan:
@taylorotwell https://jsperf.com/jquery-vs-sizzle-performance-test

taylorotwell:
So why don't we just update the jQuery code to do jQuery.find()

PR #36, opened 2017-01-18
make screenshot dir configurable

kbond:
I have been playing with integrating dusk into a Symfony app. I had to extend `Browser` to customize the screenshot directory. Not a huge deal but would be nice if it was configurable.


taylorotwell:
Merged.

PR #43, opened 2017-01-20
Make waitUntilMissing work as expected.

tabennett:
This is a PR to fix (what I consider at least) to be an issue with the `waitUntilMissing()` method in the **WaitsForElements** trait. From using the `assertMissing()` assertion in the **MakesAssertions** trait, I had assumed that this method would work the same way. That is, if I remove an element from the DOM using some JS or something, I can use this method to wait until it's gone. However this doesn't seem to be the case. If you are removing an element from the DOM and you attempt to use the `waitUntilMissing()` method, it's never going to work. `WaitUntilMissing()` is using `findOrFail()` to find an element and then check if it's visible or not. But if the element is being removed (not hidden) as you would expect when trying to use this method, the tests will always fail because the FB web driver **HttpCommandExecutor** class will throw a **NoSuchElementException** that then gets silently caught (I assume) by **WebDriverWait** class. This is pretty maddening to deal with. So what I've done here is this:

* Refactored the `waitUntilMissing()` method to use a try/catch block to catch this error (now implemented similar to the `assertMissing()` method).
* Implemented a `waitUntilHidden()` method that is essentially just the old `waitUntilMissing()` method renamed (It checks to see if an existing DOM element has been hidden).

I think this is a much consistent way to think about this stuff and is (to me at least) more correct semantically and less confusing all around.

- Made the waitUntilMissing() method work as expected (i.e it now detects when an element has been removed).
- Implemented a waitUntilHidden() method that can be used to detect when an element has been hidden.

### DIFF
--- a/src/Concerns/WaitsForElements.php
+++ b/src/Concerns/WaitsForElements.php
@@ -39,20 +39,6 @@ trait WaitsForElements
     }
 
     /**
-     * Wait for the given selector to be not visible.
-     *
-     * @param  string  $selector
-     * @param  int  $seconds
-     * @return $this
-     */
-    public function waitUntilHidden($selector, $seconds = 5)
-    {
-        return $this->waitUsing($seconds, 100, function () use ($selector) {
-            return ! $this->resolver->findOrFail($selector)->isDisplayed();
-        });
-    }
-
-    /**
      * Wait for the given selector to be removed.
      *
      * @param  string  $selector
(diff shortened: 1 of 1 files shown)


taylorotwell:
Could we just keep only waitUntilMissing but just make it work how you propose? Wait until its either not on the page at all or is not displayed?

taylorotwell:
No it's not clear because assertMissing asserts for hidden OR totally missing.

KKSzymanowski:
You're correct, my bad.

Nevertheless I feel like **missing** should mean **it's not there at all** and **hidden** should mean **it's just not visible**. At the end of the day, it's of course your decision.

taylorotwell:
@tabennett can you tweak it to just have waitUntilMissing and make it behave like assertMissing?

tabennett:
@taylorotwell So you're saying essentially we'll combine these two. So wrap the `isDisplayed()` call in a try/catch for that exception so that we can handle them both and just toss the `waituntilhidden()` method entirely? If that's what you want I can do that.

taylorotwell:
Yes

tabennett:
Done

taylorotwell:
Thanks.

PR #48, opened 2017-01-23
[WIP] Support browser kit in Dusk

sileence:
While testing JS is super awesome! It's not as fast as our old friend PHP browser kit.

That's why I'm trying to add support for Browser kit in Dusk so the developer can write and run fast tests when JS is not required/crucial. For example for static or simple pages.

The idea is to support most of the methods and the same API in the browser kit. 

The browser kit is a test response decorator similar to the one included in 5.4 but with the crawler and more methods.

My idea for the usage so far is this:

Full web driver:

```
    $browser->visit(...)->assertSee(...); 
});
```

Browser kit:

`$this->visit(...)->assertSee(...)`

This is just a WIP, as most of the methods are not working at the moment.

### DIFF
--- a/src/BrowserKit/Concerns/InteractsWithPages.php
+++ b/src/BrowserKit/Concerns/InteractsWithPages.php
@@ -1,0 +1,663 @@
+<?php
+
+namespace Laravel\Dusk\BrowserKit\Concerns;
+
+use Closure;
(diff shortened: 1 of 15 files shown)


taylorotwell:
Just please release it as a package. Sorry if I didn't make that clear.

PR #58, opened 2017-01-26
Force config:cache for enviroment file change

erikfig:
The .env.dusk or .env.dusk[enviroment] do not create a new file and by default the .env when I ran php artisan dusk.

I'm using Fedora v24. This PR fixed my problem so I think  could solve it to others.


taylorotwell:
Don't have a cache configuration file when trying to run your tests.

Issue #57, opened 2017-01-26
Explicit environment whitelisting


splatEric:
Not sure if this is the best place for this, but I wanted to make the suggestion that the Dusk login route only be exposed for specific environments.

Whilst the docs do point out that you should only include the service provider in local or testing, I wonder whether it would be better if Service Provider itself was checking this as well. To ensure full flexibility, there could be a white listing environment variable so if people wanted to expose it in other environments as well they could.

My reasoning for this is that it would be very simple for someone to skim read the docs and not pick up on this detail. If they simply included the service provider as standard, this would be something of a security flaw.

Something like:

```
class DuskServiceProvider extends ServiceProvider
{

    public function boot()
    {
        if (in_array(env('APP_ENV'), array_merge(['local','testing'], explode(',', env('DUSK_WHITELIST', ''))), false)) {
            Route::get('/_dusk/login/{userId}', [
                'middleware' => 'web',
                'uses' => 'Laravel\Dusk\Http\Controllers\LoginController@login'
            ]);

            $this->app->booted(function () {
                $this->makeLogoutAccessibleViaGet();
            });
        }
    }
```

Modelizer:
I think this would go to https://github.com/laravel/internals :)

ninthspace:
The [installation docs mention updating AppServiceProvider](https://laravel.com/docs/master/dusk#installation) to detect the environment before including the route. This is better than modifying the internals because it allows developers to **explicitly** define the environments once there rather than relying on another environment variable.

```php
 if ($this->app->environment('local', 'testing')) {
     $this->app->register(DuskServiceProvider::class);
 }
```

splatEric:
One final point to make, then I promise I will shut up about it. 

With most errors of configure, the system either doesn't work, or there's an awkward side effect such as log files filling up. If you get this one wrong, you will be opening up a hole in your authentication layer without any clue that you've done it.

SebastianS90:
@JeffreyWay Are you aware that you [present (at 0:40)](https://laracasts.com/series/whats-new-in-laravel-5-4/episodes/9) a nice way to open a huge security hole in production? Just browse to example.com/_dusk/login/1 and visit another route afterwards. You will be logged in as the user with id 1 :warning:
That works with any id, but usually id=1 is more powerful then another random user.

@taylorotwell isn't this the best example why we should protect developers from a silly mistake that has a tremendous impact?

taylorotwell:
Sure, if someone wants to submit a PR for it I will take a look.

PR #60, opened 2017-01-27
Use an authentication token for Dusk login

splatEric:
As noted in issue #57, I am concerned about the possibility of exposing the dusk login controller in an inappropriate environment and thereby having a security hole out in the wild. To this end, I have implemented an extension to the login process, where the ```InteractsWithAuthentication``` trait will store a token on the file system that the ```LoginController``` can verify for the given user.

The rationale behind this is that the token is generated in the backend, and thereby is not accessible for any would be attacker. Additionally, the token store is removed after use so that it cannot be accidentally left on a filesystem.

I had planned to use the cache, but testing revealed that the cache file was not being written by the test process until after the request was made, so this wasn't appropriate for sharing with the web server process. The additional advantage of using the file approach is that we are not dependent on the cache system not being configured for using an in memory scheme.

There may be better approaches to this (I was wondering about injecting an appropriate session variable into the browser request), but this seemed like the least invasive approach I could take that would mitigate the risk I am concerned with.

### DIFF
--- a/src/Http/Controllers/LoginController.php
+++ b/src/Http/Controllers/LoginController.php
@@ -18,11 +18,9 @@ class LoginController
      */
     public function login($userId)
     {
-        \Log::debug('here');
         if (!$this->verifyAuthToken($userId, request('token'))) {
             throw new AuthenticationException('Invalid Dusk Token for userId');
         }
-        \Log::debug('somehow');
 
         $model = config('auth.providers.users.model');
 
--- a/src/Http/Controllers/LoginController.php
+++ b/src/Http/Controllers/LoginController.php
@@ -2,18 +2,28 @@
 
 namespace Laravel\Dusk\Http\Controllers;
 
+use Illuminate\Auth\AuthenticationException;
+use Laravel\Dusk\Concerns\ManagesAuthenticationTokens;
 use Illuminate\Support\Facades\Auth;
 
 class LoginController
 {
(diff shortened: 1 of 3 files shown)


src/Http/Controllers/LoginController.php
```diff
@@ -2,18 +2,28 @@
...
     /**
      * Login using the given user ID / email.
      *
      * @param  string  $userId
      * @return Response
+     * @throws AuthenticationException
      */
     public function login($userId)
     {
+        \Log::debug('here');
```
spyric: I think it should be removed.

src/Http/Controllers/LoginController.php
```diff
@@ -2,18 +2,28 @@
...
+        if (!$this->verifyAuthToken($userId, request('token'))) {
+            throw new AuthenticationException('Invalid Dusk Token for userId');
+        }
+        \Log::debug('somehow');
```
spyric: This too

splatEric:
Yeah, I removed those in my second commit, sorry about that!

taylorotwell:
Just only register the provider in the appropriate environments as stated in the docs. Developers are able to do all sorts of bad things if they don't read the docs (leave DEBUG to true in production, etc.)

splatEric:
sure, this just seemed like something that it might be better to be a little more cautious about, given the risk. but obv. your call ;-)

PR #64, opened 2017-01-27
Fixed missing semicolon.

ecodrutz:
There seems to be a problem with IDE and parsing the information when there was no semicolon ending the use statements. It seems when it's split in two use statements the problem is gone.


taylorotwell:
File that with the IDE maintainers.

Issue #63, opened 2017-01-27
In windows 10, does laravel dusk always open the 'chrome' browser?


josealgardataborges:
Just a quick question, in windows 10, does laravel dusk always open the 'chrome' browser? I see the dusk tests like if i was using the browser. i see the input getting filled in, with username, password, for instance. Isn't it supposed to do all this stuff 'hidden' from view, just expecting the console output? It gets worse, when i use 2 browser tests, the computer almost comes to a halt. very very slow, like 30 seconds for a simple login test.

josealgardataborges:
and eventually i get this: http://prntscr.com/e11ns9

alex-LE:
I get the same timeout error. It seems that the Symfony ProcessBuilder has a default timeout of 60s.
My test is currently running > 70s and so i get the timeout. If is set the timeout to a higher value, it works.

Even if i split my tests into smaller classes/files i get the timeout.

Dusk might set a higher timeout by default, or let the user define it.

KayLeung:
Either Chrome or Firefox is not a headless browser. You could try PhantomJS or wait to see if anyone bind it to nighmare.js

deleugpn:
Can this be closed?

josealgardataborges:
@deleugpn If it's a feature not a bug, please do close it.

deleugpn:
It is indeed a feature. Not only that, it's the core concept of Dusk. The goal is to run the tests on an actual browser to have a better sense of the behavior of your application. It is possible to run Dusk on a headless machine using Xvfb. My CentOS can run a simple login test in a couple of seconds with 1 CPU Core and 1500MB of RAM.

PR #67, opened 2017-01-27
disabled process timeout

alex-LE:
set process timeout to 0 to disable it - see #63


taylorotwell:
Should be null according to doc block on method.

alex-LE:
Oh ok, thanks!

PR #72, opened 2017-01-29
DocBlock improvements

mnabialek:


taylorotwell:
ChromeDriver is one word.

https://sites.google.com/a/chromium.org/chromedriver/

mnabialek:
@taylorotwell So it's still inconsistent - in the same file there is "Chrome driver" and "Chromedriver" :) Should I change it the other way in this PR? :)

PR #79, opened 2017-01-29
Reload environment to use valid environment variables in tests

mnabialek:
At the moment I think it's working the wrong way when you have custom dusk .env file.

For example if you define in `.env` file 

```
APP_URL=http://abc.app 
```

and in `.env.dusk.local` you will define:

```
APP_URL=http://def.app 
```

Dusk will try to open `http://abc.app` url and not `http://def.app`. Also other settings will be used from original `.env` file and not from  `.env.dusk.local` so in case you use any of them in assertions, assertions will fail because tests and app is using now 2 different .env files in such case

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Dusk\Console;
 
+use Dotenv\Dotenv;
 use Illuminate\Console\Command;
 use Symfony\Component\Finder\Finder;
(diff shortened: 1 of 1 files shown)


taylorotwell:
I can't confirm this bug on my machine. If my .env is local and my .env.dusk.local has a different URL it uses the URL in the customized file. If you can provide maybe a different approach or way to recreate this feel free to open another PR.

taylorotwell:
Nevermind I recreated it.

PR #94, opened 2017-02-01
Update way of saving .gitignore content

mnabialek:


taylorotwell:
Is there a problem with the current way?

taylorotwell:
Going to leave as-is if not broken.

PR #95, opened 2017-02-01
Fix DuskTestCase not found in upgraded Laravel installation

SubodhDahal:
If Laravel is upgraded from previous version in a project, running `php artisan dusk` would throw the following error:

    PHP Fatal error:  Class 'Tests\DuskTestCase' not found in /var/www/html/projectname/tests/Browser/ExampleTest.php on line 9

Even though the file `DuskTestCase` exists, the issue is that `CreatesApplication` trait which is used in the class wasn't present in previous versions of Laravel. So if it doesn't exists then create it.

Also in the `composer.json` file, the following needs to be present.

    "autoload-dev": {
        "psr-4": {
            "Tests\\": "tests/"
        }
    },

I think that should be mentioned in the Laravel upgrade guide or Dusk installation.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -45,6 +45,10 @@ class InstallCommand extends Command
             mkdir(base_path('tests/Browser/screenshots'), 0755, true);
         }
 
+        if (! file_exists(base_path('tests/CreatesApplication.php'))) {
+            copy(__DIR__.'/../../stubs/CreatesApplication.php', base_path('tests/CreatesApplication.php'));
+        }
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
I believe this is alone will not really help since probably people won't have the composer.json file updated. A better solution is just to improve the upgrade guide in this area.

SubodhDahal:
Yes, this alone won't fix the issue. Having to update the `composer.json` file definitely needs to be added to the upgrade guide. But having this commit will mean one less step for the user to perform for the upgrade.

PR #103, opened 2017-02-06
Add assertQueryIs and assertFragmentIs

KKSzymanowski:
This provides two more assertions for the URL.
- `assertQueryIs`
- `assertFragmentIs`

Using `parse_url` function, an example URL `http://app.dev/foo?bar=baz#qux` will be split up into
```php
[
  "scheme" => "http"
  "host" => "app.dev"
  "path" => "/foo"
  "query" => "bar=baz"
  "fragment" => "qux"
]
```
Currently existing `assertPathIs` method allows only for assertion of the `path` substring and there may be situations where we need to check for other parts of the URL.

If the `query` part is not present, it is treated as an empty string, which allows for following:
```php
$browser->assertQueryIs('');
// or
$browser->assertQueryIs(null);
```
when we need to check if the query is empty. Same with `fragment`.

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -50,6 +50,45 @@ trait MakesAssertions
 
         return $this;
     }
+    
+    
+    /**
+     * Assert that the current URL query matches the given query.
+     *
+     * @param  string  $query
+     * @return $this
+     */
+    public function assertQueryIs($query)
+    {
+        $url = parse_url($this->driver->getCurrentURL());
+
+        if(!isset($url['query'])) {
+            $url['query'] = '';
+        }
+
+        PHPUnit::assertEquals($query, $url['query']);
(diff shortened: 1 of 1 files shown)


taylorotwell:
Browser is Macroable to allow for easy adding of assertions.

KKSzymanowski:
@taylorotwell Thanks, I haven't thought about that.

ryantbrown:
Found this when I was looking for a way to test an app with a lot of JS routing. Just my 2 cents but it would be great to have `assertFragmentIs` and `assertFragmentHas`. Creating a macro is fine but considering it's the only way to test a single page app/section without push I think it warrants a native method.

I created the macro for

KKSzymanowski:
I agree although I understand Taylor's reasoning. If this gets more support from the community I can resubmit the PR.

Issue #102, opened 2017-02-06
Change extension of stubs


KKSzymanowski:
Currently dusk stubs have the `.php` extension which causes PhpStorm(maybe other IDEs) to warn about multiple class definitions for eg. `DuskTestCase`. 
Can we change the extension to `.stub`, like for generator commands?

This will require changes in `Laravel\Dusk\Console\InstallCommand::handle`:
Before:
```php
copy(__DIR__.'/../../stubs/ExampleTest.php', base_path('tests/Browser/ExampleTest.php'));
copy(__DIR__.'/../../stubs/HomePage.php', base_path('tests/Browser/Pages/HomePage.php'));
copy(__DIR__.'/../../stubs/DuskTestCase.php', base_path('tests/DuskTestCase.php'));
copy(__DIR__.'/../../stubs/Page.php', base_path('tests/Browser/Pages/Page.php'));
```
After:
```php
copy(__DIR__.'/../../stubs/ExampleTest.stub', base_path('tests/Browser/ExampleTest.php'));
copy(__DIR__.'/../../stubs/HomePage.stub', base_path('tests/Browser/Pages/HomePage.php'));
copy(__DIR__.'/../../stubs/DuskTestCase.stub', base_path('tests/DuskTestCase.php'));
copy(__DIR__.'/../../stubs/Page.stub', base_path('tests/Browser/Pages/Page.php'));
```

What do you think?

PR #109, opened 2017-02-08
Rename stubs to avoid multiple class definition warning in PHPStorm

sileence:
PHPStorm (and I'm guessing other IDEs as well) is complaining because there are multiple definitions of DuskTestCase basically because of the stubs, it's really annoying and I don't want to use Sublime. Could you please merge this one? 

Fixes https://github.com/laravel/dusk/issues/102


KKSzymanowski:
@sileence Thanks.

PR #112, opened 2017-02-09
Accept relative or absolute URLs in the assert path is method

sileence:
This change makes the `assertPathIs` method more versatile: 

```
                ->assertPathIs("/post/{$post->slug}")
                ->assertPathIs("post/{$post->slug}")
                ->assertPathIs("http://forum.app/post/{$post->slug}")
                ->assertPathIs(route('posts.show', [$post->slug]));
```

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -44,9 +44,10 @@ trait MakesAssertions
      */
     public function assertPathIs($path)
     {
-        PHPUnit::assertEquals($path, parse_url(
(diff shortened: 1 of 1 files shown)


taylorotwell:
Would rather just accept relative URLs.

sileence:
@taylorotwell `visit` accepts absolute or relative URLs. It seems strange that you can visit any URL but assert paths of relative URLs only.

BTW I found some room for refactoring if you want to reconsider this feature.

PR #114, opened 2017-02-09
Add assert code has and assert code missing methods

sileence:
I have some tests with the old browser kit to make sure the content of a post was converted from markdown to valid and escaped HTML. So I tried to reproduce the test with Dusk:

```
            $browser->visit($post->url)
                ->assertSee($xssAttack)  // see the plain text
                ->assertCodeMissing($xssAttack); // but not the HTML code (because it was escaped)
```

I'll leave this PR in case you find it useful.

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -164,6 +164,34 @@ trait MakesAssertions
     }
 
     /**
+     * Assert that the given source code is present on the page.
+     *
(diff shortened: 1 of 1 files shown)


MightySCollins:
I would really like this for some of my tests but would it make more sense to be `assertSourceHas`?

sileence:
done, thank you for the feedback @MightySCollins

PR #118, opened 2017-02-10
Store the HTML output of the failed tests.

sileence:
This PR creates an output folder inside Browser/ that will store the HTML of the failed tests (similar to the screenshot feature).

I think having the output together with the screenshot will be very useful to debug those failing tests :)

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -36,6 +36,13 @@ class Browser
     public static $storeScreenshotsAt;
 
     /**
+     * The directory that will contain any source codes.
+     *
+     * @var string
+     */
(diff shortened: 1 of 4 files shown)


taylorotwell:
Don't really want to go this far.

deleugpn:
This could have been a great asset. People running Dusk on headless VM can't leverage `stop()` to check the browser. The HTML could be the way out.

slickorange:
Agree with @deleugpn 
Struggling with a test where assertion fails, but I can see the element is there on the screenshot... (Running tests headless on Homestead)

Issue #97, opened 2017-02-02
[Request] Proposal for Browser class


brandonferens:
All of the stubs reference the Browser class at `Laravel\Dusk\Browser`. My proposal is that a Browser class is installed within the tests that extends `Laravel\Dusk\Browser`. Not exactly sure the best/most logical place for this file, perhaps `tests/Dusk/Browser.php`.

The reason for this is that I have created some custom methods on the browser like `loginAsUser()` and `loginAsAdmin()`. I also set those users into the object and are made available as `$browser->user`. This is just part of my use case, but I can see the potential for people needing this kind of flexibility on their browser object.

If you would like, I can create a PR with the necessary code to accomplish this feature.

brandonferens:
Thanks @sileence, that is very close to what I created in my DuskTestCase!

deleugpn:
Should this be closed?

browner12:
no, i think his is different. 3b2e5c4 is just a little optimization.

I'm all for this update, for the exact use cases you suggested.

brandonferens:
No, this should be closed. 3b2e5c4 is a much better approach than what I came up with. You can now add something like this to your `DuskTestCase`:

    /**
     * Create a new Browser instance.
     *
     * @param  \Facebook\WebDriver\Remote\RemoteWebDriver  $driver
     * @return \Laravel\Dusk\Browser
     */
    protected function newBrowser($driver)
    {
        return new MyOwnBrowser($driver);
    }

This would then override the default class and would load your custom browser class.

Within that custom browser class, you could have a method like this:

    /**
     * @var int
     */
    protected $adminId = 1;

    /**
     * @var \App\Models\User
     */
    public $user;

    /**
     * Log into the application using a given user ID or email.
     *
     * @return $this
     */
    public function actingAsAdmin()
    {
        $this->user = User::find($this->adminId);

        return $this->loginAs($this->user);
    }

This allows you to do something like this within your dusk test

    $browser->actingAsAdmin()->visit('/admin/page');

browner12:
ahh, got it now.  The intent is that you override the `newBrowser` method if you want a custom browser.

brandonferens:
Exactly.

PR #119, opened 2017-02-10
Add the possibility to extend the browser class.

sileence:
With this change it is now possible to extend the browser class by overriding the `newBrowser` method in the `DuskTestCase` class:

```
    /**
     * Create a new Browser instance.
     *
     * @param  \Facebook\WebDriver\Remote\RemoteWebDriver  $driver
     * @return \Laravel\Dusk\Browser
     */
    protected function newBrowser($driver)
    {
        return new MyBrowser($driver);
    }
```

`class MyBrowser extends Browser`

Possible solution to: https://github.com/laravel/dusk/issues/97

I know the Browser class is macroable but sometimes it is better to just extend it, especially when you want to add a lot of new functionality.

### DIFF
--- a/src/TestCase.php
+++ b/src/TestCase.php
@@ -108,16 +108,27 @@ abstract class TestCase extends FoundationTestCase
     protected function createBrowsersFor(Closure $callback)
     {
         if (count(static::$browsers) === 0) {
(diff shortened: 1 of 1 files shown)


deleugpn:
Just replaced my macros with this and it is perfect!

PR #122, opened 2017-02-13
Split Arguments for PHPUnit and merged with defaults

MightySCollins:
I originally made #115 and somewhat incorrectly used an array union which seemed to work but actually did not...

What I should have done done is used an `array_merge` like it did originally. I have tried this with `--group foo` and below is the output before the command is run by the `Process` class:

    string(98) "'vendor/bin/phpunit' '-c' 'PROJECT_ROOT/phpunit.dusk.xml' '--group' 'foo'"

Do you think its worth making a test with a mocked ProcessBuilder?

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -45,7 +45,7 @@ class DuskCommand extends Command
     {
         $this->purgeScreenshots();
 
-        $options = implode(' ', array_slice($_SERVER['argv'], 2));
+        $options = array_slice($_SERVER['argv'], 2);
 
         return $this->withDuskEnvironment(function () use ($options) {
             return (new ProcessBuilder())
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you let me know what this is fixing?

MightySCollins:
So this is fixing a problem I have with providing custom arguments to dusk. Currently from what I can tell custom options are imploded into a string and added as a third item in the array. This can be a problem because it means commands seem to be run like below:

    'vendor/bin/phpunit' '-c' 'PROJECT_ROOT/phpunit.dusk.xml' '--group foo --log-junit dusk.log'

When this runs PHPUnit returns no such option as it parses the entire last parameter as one thing.

There is slightly more detail in #115 but this change should split the arguments out so each parameter is wrapped in `'` by the process builder.

PR #128, opened 2017-02-15
Make sure the user is always logged out after the test finishes

sileence:
I noticed an issue involving tests with authentication.

If I logged in the user in my first test, like so:

```
            $browser->loginAs($this->user());
        });
```

It remains authenticated in the following tests because the main browser remains open (`closeAllButPrimary`). These of course can create some false positives, etc. 

This is why I'm adding this flag to make sure the user is log out after the test is completed. So the next test can begin with a clean state.

### DIFF
--- a/src/Concerns/InteractsWithAuthentication.php
+++ b/src/Concerns/InteractsWithAuthentication.php
@@ -7,6 +7,13 @@ use Laravel\Dusk\Browser;
 trait InteractsWithAuthentication
 {
     /**
+     * Indicates if the user was authenticated using the loginAs method.
+     *
+     * @var bool
+     */
+    protected $loggedIn = false;
+
+    /**
      * Log into the application as the default user.
      *
      * @return $this
@@ -26,6 +33,8 @@ trait InteractsWithAuthentication
     {
         $userId = method_exists($userId, 'getKey') ? $userId->getKey() : $userId;
 
+        $this->loggedIn = true;
+
         return $this->visit('/_dusk/login/'.$userId);
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I'm not sure I want to build this assumption into Dusk.

sileence:
@taylorotwell it's not an assumption, but the fact that if a user is authenticated in any test method of the test suite, from that point on it will be authenticated in all the other tests.

We don't even control the order in which the tests are executed, so we cannot rely on this behaviour,  and the tests should produce the same result whether they are run together or individually.

Example:

```
    function test_welcome_known_user()
    {
            $browser->loginAs(user('Duilio'))
                  ->visit('welcome')
                  ->assertSee('Welcome Duilio!');
        });
    }

    function test_welcome_guest_user()
    {
            $browser->visit('welcome')
                 ->assertSee('Welcome Guest!');
        });
    }
```

Executing `php artisan dusk` -> `test_welcome_guest_user` will fail because the user is already authenticated from `test_welcome_known_user`

`php artisan dusk --filter= test_welcome_guest_user ` will pass because the user is not authenticated.

If we inverse the order of the methods:

```
    function test_welcome_guest_user()
    {
        //...
    }

    function test_welcome_known_user()
    {
        //...
    }
```

`php artisan dusk` -> will pass, because `test_welcome_guest_user` is executed first.

Manual fix: call `logout` at the end or beginning of certain tests:

```
    function test_welcome_known_user()
    {
            $browser->loginAs(user('Duilio'))
                  ->visit('welcome')
                  ->assertSee('Welcome Duilio!')
                  ->logout();
        });
    }

    function test_welcome_known_user()
    {
        //...
    }
```

`php artisan dusk` -> will pass, because `test_welcome_known_user ` logouts the user at the end of the test.

***

Maybe my solution is not ideal but this is an issue that I think it is important to consider.

Another solution could be to just close all the browsers but I'm assuming this will make the tests slower? Or maybe having another way to indicate if we want to run the tests with a known user or not.

PR #129, opened 2017-02-15
Provide custom logout route

sileence:
I ran into another issue because in my app the logout's route is post and not get, therefore the following call: `return $this->visit(route('logout', [], false));`  was throwing an error instead of the logout action.

I think Dusk should provide its own logout action to avoid that issue, also to avoid assuming that the logout's route is defined with the "logout" name, etc.

### DIFF
--- a/src/Concerns/InteractsWithAuthentication.php
+++ b/src/Concerns/InteractsWithAuthentication.php
@@ -36,6 +36,6 @@ trait InteractsWithAuthentication
      */
     public function logout()
     {
-        return $this->visit(route('logout', [], false));
+        return $this->visit('/_dusk/logout/');
     }
 }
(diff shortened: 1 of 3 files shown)


sileence:
I just noticed the `makeLogoutAccessibleViaGet` code, but it is not working for me, I think it is because I have this route:

```
// All posts
$router->get('{category?}', [
    'uses' => 'ListPostController@index',
    'as'   => 'posts.index',
]);
```

Apparently this is being registered before the `makeLogoutAccessibleViaGet` method is executed and therefore causing the error. I think my solution on this issue might be better, because with `logout()` we are not trying to test an action but just to logout the user (same as `loginAs`).

There might be other scenarios where the logout action is incomplete or broken but we are not trying to test that as long as I see it.

PR #132, opened 2017-02-17
Bug in formatting elements with similar names

thijsvdanker:
## Problem
Given:

- there are two page elements
- the key of the second element contains the key of the first element

```php
[
  '@foo' => '#first',
  '@foobar' => '#second'
]
```
When:

- the format method replaces the value

Then:

- the second element is replaced in part with the first elements value `#firstbar` instead of `#second`

## Solution
I'm not sure how to solve this, `preg_replace` might be the proper solution?
PR provides a failing test.


taylorotwell:
@themsaid can you take a look at this?

themsaid:
Thank you, that was fixed in https://github.com/laravel/dusk/pull/135

PR #139, opened 2017-02-20
Log console content into files

themsaid:
This PR adds the functionality of storing the browser console content into files for further inspection. Upon installation a `Browser/console` directory will be created to store the log files of each test.

An example log file looks like:

```
[
    {
        "level": "SEVERE",
        "message": "http:\/\/duskproject.dev\/ 18:28 Uncaught TypeError: Cannot read property 'undefined' of null",
        "source": "javascript",
        "timestamp": 1487615364745
    }
]
```

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -36,6 +36,13 @@ class Browser
     public static $storeScreenshotsAt;
 
     /**
+     * The directory that will contain any console log.
+     *
+     * @var string
+     */
+    public static $storeConsoleLogAt;
+
+    /**
      * Get the callback which resolves the default user to authenticate.
      *
      * @var \Closure
@@ -197,6 +204,22 @@ class Browser
     {
         $this->driver->takeScreenshot(
             sprintf('%s/%s.png', rtrim(static::$storeScreenshotsAt, '/'), $name)
+        );
+
+        return $this;
+    }
+
+    /**
+     * Store the console output with the given name.
+     *
+     * @param  string  $name
+     * @return $this
+     */
+    public function logConsole($name)
+    {
+        file_put_contents(
+            sprintf('%s/%s.log', rtrim(static::$storeConsoleLogAt, '/'), $name)
+            , json_encode($this->driver->manage()->getLog('browser'), JSON_PRETTY_PRINT)
         );
 
         return $this;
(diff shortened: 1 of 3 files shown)


devinfd:
:)

SebastianS90:
That looks really nice! I think it would be useful to have the logs accessible within your tests via a fluent API and already before `::browse()` finishes. For example:
```php
$browser
    ->visit('/')
    ->doSomething()
    ->assertLogEmpty()
    ->doSomethingElseThatLogsMessages()
    ->assertLogContains('this is a message I expect to be in the browser log');
```

Javascript frameworks like `vue` print log messages when you are using them in a wrong way (only enabled when compiling with env != production). Using such assertions can help to catch that in your test suite.

Just an idea, feel free to ignore it :laughing:

SeoFood:
could you add a .gitignore file same as the screenshot dir? Because if I just check out the Project I get an error:


```Browser/console/testBasicExample-0.log): failed to open stream: No such file or directory```

**Comment by bbashy** [comment]
Same thing. Need to create folder automatically?
```
1) Tests\Browser\Admin\UserTest::canCreateUser
ErrorException: file_put_contents(/path/to/project/tests/Browser/console/testExample-0.log): failed to open stream: No such file or directory
```

sileence:
@taylorotwell since this was merged, is there any possibility you could reconsider storing the HTML output as well? https://github.com/laravel/dusk/pull/118

(I could rework the PR to make the code better, etc.)

SebastianS90:
@bbashy It's created when running `php artisan dusk:install` but not when upgrading dusk to a new version.
But running `dusk:install` again is a bad idea because [it will overwrite](https://github.com/laravel/dusk/blob/v1.0.9/src/Console/InstallCommand.php#L52-L59) some files you might have changed.

We should wrap these commands in `if (! file_exists(...))` and then we can recommend to run `dusk:install` again after an upgrade.

SeoFood:
yeah and there is no `.gitignore` file in the console dir.

SebastianS90:
When #166 is merged and tagged and you have upgraded dusk to the new version, then you can just run `php artisan dusk:install` again and it will create the folder together with a `.gitignore` file for you without overwriting existing files.

Until then, you should just create the `tests/Browser/console` folder and the `tests/Browser/console/.gitignore` file manually.

rogervila:
I think dusk should create the `console` directory with the `.gitignore` file if they are missing when it runs.

https://github.com/laravel/dusk/issues/167

SebastianS90:
@rogervila sounds even better than my PR. And it might also be good to purge old logs on each run (as we already do with screenshots)

PR #141, opened 2017-02-21
Add user info to the response

sileence:
I was trying to find a way to get the user info (and possibly other useful data) from the response itself, in that way `assertAuthenticated*` methods won't need to send another request and the page is not changed, i.e. this is now possible: 

```
    $browser->visit($token->url)
        ->assertAuthenticatedAs($user) // the page is not changed, if this fails the proper screenshot will be captured
        ->assertPathIs('/'); // and also you can continue calling the assert* methods
```

This is probably not the best solution but I'll leave the PR here just in case :)


taylorotwell:
I don't really want to do this.

Issue #142, opened 2017-02-21
loginAs fails with SessionGuard::login() error when called for a second user.


danielbaylis:
When you call `loginAs` for a second user in a second test after a _different_ user has already been logged in in an earlier test, an exception is thrown, the second user is not logged in, and the test continues to run as the first user.

The test doesn't fail despite the exception in the second `loginAs` call and no screenshot is captured.

I realize there is a `logout` method we can call on $browser - so perhaps it's just a matter of doing this at the end of each test.  But would it be better/safer if `loginAs` could also logout any existing user?

The concern is the second user never gets logged in, so even though a test might pass it hasn't actually run on the correct user. Tests still return green so it might not be obvious this has occurred unless you are actually watching the (high speed) chrome window that pop-ups during test execution and notice the exception page flash up.

Dusk is a really awesome tool - Thanks!

## Exception
```
Symfony\Component\Debug\Exception\FatalThrowabl... ·GET /_dusk/login/101396
Type error: Argument 1 passed to Illuminate\Auth\SessionGuard::login() must implement interface Illuminate\Contracts\..
```

## Example
Just as an example, here are two tests that log in different users and keep the browser open for a couple of seconds so I can visually confirm in my application that the user hasn't changed.  I am needing to login in as specific users to test against a specific existing data set.
```
    public function testFirstUser()
    {
        $user = User::find(1);

        $this->browse(function ($browser) use ($user) {
            $browser->loginAs($user)
                ->visit('/')
                ->pause(2000);
        });
    }

    public function testSecondUser()
    {
        $user = User::find(1000);

        $this->browse(function ($browser) use ($user) {
            $browser->loginAs($user)
                ->visit('/')
                ->pause(2000);
        });
    }
```

sileence:
I was talking about this issue (and trying to solve it) here: https://github.com/laravel/dusk/pull/128 

For now I'm just calling `->logout()` at the end of all the tests that use `loginAs`

danielbaylis:
Thanks for the info @sileence - I can see how some users might like the convenience of having the logged in user _roll on_ through the tests. Logging them out automatically after each test would take away our ability to choose whether to log out them out, and could be a big breaking change for all the tests already written.  I think @taylorotwell means he doesn't want dusk to assume we want the user logged out after each test - we should have the choice. 

I see what you mean about the tests having different results when executed alone or as a group though.

As for my issue, it was actually a global scope issue causing the second user not to be found. 
[...]

PR #143, opened 2017-02-21
Delete all the cookies to logout the authenticated user

sileence:
Another possible and simpler solution to this issue: https://github.com/laravel/dusk/issues/142 

Also described here: https://github.com/laravel/dusk/pull/128

The idea is to delete all the cookies so the session is reset and the next test begins with a clean state (no auth user, etc.)

### DIFF
--- a/src/Concerns/InteractsWithCookies.php
+++ b/src/Concerns/InteractsWithCookies.php
@@ -85,4 +85,14 @@ trait InteractsWithCookies
 
         return $this;
     }
+
+    /**
+     * Delete all the cookies.
+     *
+     * @return $this
+     */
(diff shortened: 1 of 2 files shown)


taylorotwell:
I still do not think we should automatically make the assumption of deleting all cookies.

themsaid:
@sileence maybe you can just logout manually after each test "if you want", that way you don't have to re-auth the user on every test if you rely on a setup that's meant to run on all tests in the class.

For example, you log a user in and test multiple scenarios in different methods.

deleugpn:
@themsaid but isn't each test suppose to be like a clean _sandbox_?

PR #144, opened 2017-02-22
Return the browser object to allow chaining after a call to ->script.

danielbaylis:
Returning the browser object allows us to execute a script (or scripts) and then continue chaining test methods.

Example use case is simulating a user scrolling to the bottom of the page (to make a button visible) and then pressing it.

We could split this to a different method if there is a need to continue returning the collection of javascript execution results as it works now.


taylorotwell:
Breaking change. We would need to continue returning JavaScript results in some cases.

PR #145, opened 2017-02-22
add selectorPrefix to clickLink method

devinfd:
With clickLink it was possible to click the wrong link if multiple links exist with the same (or similar) content. $selectorPrefix allows user to target with stronger specificity.


taylorotwell:
Use the ->with() method then call ->link()

devinfd:
Nice. Missed that in the docs. Thanks!

Devin Dombrowski
dddesigns3@gmail.com

PR #147, opened 2017-02-22
Support for contextClick (AKA right click) actions

CJLees01:


taylorotwell:
Why not just call it rightClick?

CJLees01:
I was following Facebook naming under the covers. 
I agree though that rightClick is more intuitive, so I'll change it.

PR #151, opened 2017-02-23
Added query string assertions

devinfd:
Added assertHasQueryStringParam, assertMissingQueryStringParam and assertQueryStringValue.


taylorotwell:
This should only be two methods.. ->assertQuery($name, $value = null)... ->assertMissingQuery($name)

devinfd:
Yea I was wondering about that. Are you sure about assertQuery/assertMissingQuery? Might sound like they are referring to sql queries.

PR #154, opened 2017-02-23
do not blindly overwrite configuration files

### LINKED ISSUE #130: phpunit.dusk.xml not preserved
Is it intentional that the `phpunit.dusk.xml` file is [always overwritten](https://github.com/laravel/dusk/blob/v1.0.6/src/Console/DuskCommand.php#L113) before running the tests and [deleted](https://github.com/laravel/dusk/blob/v1.0.6/src/Console/DuskCommand.php#L116) afterwards?

### LINKED ISSUE #153: .env is overridden with .env.dusk if Dusk doesn't stop cleanly
See [this comment](https://github.com/laravel/dusk/issues/152#issuecomment-282120330) by @georaldc

- Have a `.env.dusk` file which is different from your `.env` file
- Start `php artisan dusk`
  - Your original `.env` will be backed up to `.env.backup`
  - Your `.env.dusk` file will be copied over the normal `.env` file
- Assume dusk somehow fails, or you simply abort it with `Ctl+C`
  - The backup is **not** restored, leaving your Dusk configuration in `.env`
- Start `php artisan dusk` again
  - It will copy the `.env` file (which contains the Dusk configuration) to `.env.backup`, thus deleting your normal environment configuration

SebastianS90:
closes #153
closes #130

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -25,6 +25,13 @@ class DuskCommand extends Command
     protected $description = 'Run the Dusk tests for the application';
 
     /**
+     * Whether the project has its own phpunit.dusk.xml file.
+     *
+     * @var boolean
+     */
+    protected $phpunitDuskXmlExists = false;
+
+    /**
      * Create a new command instance.
      *
      * @return void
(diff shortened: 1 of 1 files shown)


taylorotwell:
Instead of all this why not just use a try / finally or something?

SebastianS90:
@taylorotwell try-finally does not always help:

`finally.php`:
```php
<?php
try {
	sleep(10);
} finally {
	echo "Finally!\n";
}
```

```bash
sebastian@seb-pc ~ % php finally.php 
^C
130 sebastian@seb-pc ~ %
```
Finally does not execute when pressing `Ctrl+C`

And finally would not help to preserve a custom `phpunit.dusk.xml` either. Can you please merge it or point me to a different solution for both bugs?

PR #156, opened 2017-02-24
Add stop() to stop the PHPunit process but keep the browser open

themsaid:
Using `stop()` will exit the process that's running PHPUnit but will not attempt to close the currently open browser since the `afterClass` is not called.

This is useful for debugging when you want to keep the browser open and tinker with elements at a given state.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -283,7 +283,7 @@ class Browser
     }
 
     /**
-     * Stop the script but keep the browser open.
+     * Stop running tests but leave the browser open.
      *
      * @return void
      */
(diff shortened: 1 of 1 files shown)


salarmehr:
@themsaid  This trick does not work for me and the browser still is closed after finishing test. Is there any special way to use it?

salarmehr:
**Notice** This dummy function does not work with ChromeDriver, use selenium if you need to keep browser open. 

Not sure if there is any support for that in ChromeDriver.

salarmehr:
^ this must mentioned some place

PR #160, opened 2017-02-27
Configuration

jaesung2061:
Makes paths configurable. Useful for non-default Laravel installations.


taylorotwell:
I don't like to add a ton of configuration options.

PR #164, opened 2017-03-02
Added drag by offset functionality

helmut:
I found a need for some offset base dragging functionality and added the following few methods to allow me to do that. I thought they may be a useful addition to the core so just submitting here in case you are interested. 

No stress if you aren't keen.

### DIFF
--- a/src/Concerns/InteractsWithElements.php
+++ b/src/Concerns/InteractsWithElements.php
@@ -326,6 +326,71 @@ trait InteractsWithElements
 
         return $this;
     }
+    
+    /**
+     * Drag an element up.
+     *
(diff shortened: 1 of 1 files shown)


taylorotwell:
What if "offset"... the number of pixels to drag in that direction?

helmut:
I didn't see it officially stated anywhere, but I assume that is the unit of measurement. Appears that way from my tests. I believe it is taken from the middle of the element you have selected. It's part of the `Facebook\WebDriver\Interactions\WebDriverActions`. View the class [here](https://github.com/facebook/php-webdriver/blob/community/lib/Interactions/WebDriverActions.php).

PR #166, opened 2017-03-02
allow to run dusk:install as upgrade

SebastianS90:
Please see the discussion in #139


bbashy:
We may want to keep the .gitignore creations outside the `if`s because then it will add the file if it exists when it may not have .gitignore.

PR #171, opened 2017-03-03
Tiny Fix, duplicate backslash

edenreich:
After a while it took me time to figure out why I always get Failed asserting that two strings are equal //contact equals /contact

failed

It appears that from where it pulls the configuration(.env file APP_URL) if you have backslash at the end you always get different strings with assertEquals:

env

when I remove that backslash at the end of the file manually the test is passeding.

so in order to make this small fix I suggest to force it to have only one backslash even if the someone adds backslash at the end in the .env file on the APP_URL variable.

now both cases passing..

success


taylorotwell:
I suggest just setting the URL correctly in the env file.

PR #176, opened 2017-03-07
Added method WaitsForElements::waitForLocation()

sahibalejandro:
Method assertPathIs() sometimes fails if your script are redirecting the user to another location after an AJAX call is completed, for example when building a SPA using Vue Router.

This method waitForLocation() solves this issue actually waiting for the window.location.pathname to change.

### DIFF
--- a/src/Concerns/WaitsForElements.php
+++ b/src/Concerns/WaitsForElements.php
@@ -95,7 +95,7 @@ trait WaitsForElements
      */
     public function waitForLocation($path, $seconds = 5)
     {
-        return $this->waitUntil("window.location.pathname == '$path'", $seconds);
+        return $this->waitUntil("window.location.pathname == '{$path}'", $seconds);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


KKSzymanowski:
The `Browser` class is macroable which means you can easily extend it's functionality. Here's what I did for *fragments* (example.com/#**/foo/bar**):

```php
Browser::macro('assertFragmentIs', function($fragment) {
    $url = parse_url($this->driver->getCurrentURL());

    if(!isset($url['fragment'])) {
        $url['fragment'] = '';
    }

    PHPUnit_Framework_Assert::assertEquals($fragment, $url['fragment']);

    return $this;
});

Browser::macro('waitUntilFragmentIs', function ($fragment, $seconds = 5) {
    return $this->waitUsing($seconds, 100, function () use ($fragment) {
        return $this->assertFragmentIs($fragment);
    });
});
```

You can put it wherever you like. I suggest creating a separate service provider if you plan to have more of these macros. If not `DuskTestCase::setUp` will be just fine.

PR #186, opened 2017-03-12
Add ability to assertSeeCollection and assertSeeCollectionIn

deleugpn:
Loop through the collection and assert each given label has a value that can be found on the page or within the given selector.


taylorotwell:
No plans on adding this to core. Use a macro.

Issue #189, opened 2017-03-13
[Feature Request] assertPathIs does not play well with pound symbol


dominiquedutra:
Hey,

Just noticed that the current available assertPathIs does not play well with javascript routing.
Some libraries (flatiron/director) add a # (pound/sharp) to the URL:

http://foo.com/url/#/bar

$browser->assertPathIs() reads this as /url and stops there.

Best regards

deleugpn:
What do you think about this kind of usability?

```
$this->browse(function (Browser $browser) use ($user) {
    $browser->visit('http://laravel-admin.herokuapp.com/#/login')
        ->assertPathIs('/')
        ->assertFragmentIs('/login');
});
```

@taylorotwell is this what you meant on #191 ?

dominiquedutra:
@deleugpn that would be perfect in my opinion.

KKSzymanowski:
@deleugpn I already tried that with #103

driesvints:
Heya. Browser class is macroable so feel free to overwrite or add methods for your use case.

PR #191, opened 2017-03-13
Allow assert path with anchor link

deleugpn:
#189 and anchor links in general.


taylorotwell:
Personally would rather have separate methods for path and fragment assertions.

PR #205, opened 2017-03-22
Require latest php-webdriver version

OndraM:
There is a bunch of useful improvements and fixes in latest php-webdriver version, see [changelog](https://github.com/facebook/php-webdriver/blob/community/CHANGELOG.md).

IMHO there is no need to rely on the ~1.0 and the 1.4+ could be forced, or am I wrong?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": ">=5.6.4",
-        "facebook/webdriver": "~1.0",
+        "facebook/webdriver": "^1.4",
         "nesbot/carbon": "~1.20",
         "illuminate/support": "~5.3",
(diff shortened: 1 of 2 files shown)


taylorotwell:
~1.0 will already install 1.4. That's the point.

OndraM:
But not necessarily. By "~1.0" you are saying "any 1.x version", though you don't tests against any lower then latest dev-dependencies of required packages, so IMO you can never be sure you are "~1.0" compatible.

For example, you may be using some feature added in 1.3. And someone depending on Dusk may have  some constrain causing Composer to resolve dependencies by installing 1.2, which will then fail the Dusk tests.

¯\\\_(ツ)_/¯

PR #214, opened 2017-04-07
Restore support for IDs containing a colon

duncan3dc:
#163 broke support for IDs containing a colon, eg:
```html
<input type='text' id='#frmLogin:strCustomerLogin_userID'>
```
This pr restores it and adds a test

### DIFF
--- a/src/ElementResolver.php
+++ b/src/ElementResolver.php
@@ -276,7 +276,7 @@ class ElementResolver
      */
     protected function findById($selector)
     {
-        if (preg_match('/^#[\w\-]+$/', $selector)) {
+        if (preg_match('/^#[\w\-:]+$/', $selector)) {
             return $this->driver->findElement(WebDriverBy::id(substr($selector, 1)));
         }
     }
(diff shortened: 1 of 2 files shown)


taylorotwell:
@JExperton any comment on this? Does this look ok?

JExperton:
Two things:

- The given example #frmLogin:strCustomerLogin_userID with leading hash in the ID attribute won't match because the hash character is not part of the "\w" class. 
Generally speaking, you shouldn't start an ID attribute with the # character because it doesn't work in a stylesheet. I guess it's just a confusion in the example between CSS selector and HTML attribute.

- Using a colon in an ID doesn't work in a stylesheet because the character is meant to be used in pseudo-selectors such as :hover or :first-child. Not a good idea to use it in an ID.

However, it looks good to me since an ID should be unique and there's no point to combine it with a pseudo-selector. Even if selectors like #element:first-child or #element:visited will match, I think it's ok in the context of Dusk.

I was thinking about replacing the regex with ^#[\S]+ to fit the W3C definition of ID and avoid future requests to add support for other characters. Then, selector such as #element[prop] and #element>child will match which is not an expected behavior but might be acceptable if we take for granted that developers know what they are doing.

The question is do we stick to patterns that work in a stylesheet or to the W3C definition of an ID?

duncan3dc:
I'm +1 on the W3C definition, looking for whitespace to end the ID

PR #218, opened 2017-04-11
add case sensitive option to assertions

browner12:
if the user wants the test to be case insensitive we will just lower-case everything.  by default assertions remain case-sensitive by default to maintain BC.

also add some tests.

I am not the world's most experienced unit tester, so suggestions are welcome.

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -2,9 +2,9 @@
 
 namespace Laravel\Dusk\Concerns;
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I would prefer not to go down this road as I feel it somewhat complicates the API and locks us in from offering any additional parameters there.

browner12:
any suggestions for other ways to tackle this then? because i think the case sensitivity can make some tests unnecessarily brittle.

PR #136, opened 2017-02-19
Implement a method "isTtySupported" for setTty method.

mamor:
I tried dusk with PhantomJS on Wercker. and I have same issue this.
https://laracasts.com/discuss/channels/testing/tty-mode-requires-devtty-to-be-writable

Like that comment "run as script" can be executed.
but Wercker continued build when dusk fails, probably exit code does not return.
This problem will be occored on CircleCI, GitLab CI, cron and more.

I fixed that, change to the same logic Symfony\Component\Process\Process#setTty()
https://github.com/symfony/process/blob/master/Process.php#L1036

and this is execution result.

```
$ cat /tmp/test.php 
<?php

var_dump(PHP_OS !== 'WINNT');

var_dump(PHP_OS !== 'WINNT' && (bool) @proc_open('echo 1 >/dev/null', [['file', '/dev/tty', 'r'], ['file', '/dev/tty', 'w'], ['file', '/dev/tty', 'w']], $pipes));
```

iTerm2

```
$ php /tmp/test.php 
bool(true)
bool(true)
```

jenkins on MacOSX

```
+ php /tmp/test.php
bool(true)
bool(false) // no supported tty even if OS is not Windows.
```

PR #223, opened 2017-04-17
Attempt to solve TTY issue

deleugpn:
Following up on #136, `is_writable` seems A LOT less harmless and simpler to check before attempting to enable TTY. However, I would need somebody actually suffering from the issue to check if this actually resolves the problem.

The only other solution would be to break the chain and catch the RuntimeException thrown by Symphony, in my opinion.


giolf:
i'm building ... i'll let you know soon ...

Issue #230, opened 2017-04-19
Package use the "illuminate/console", but it is not listed as dependency of composer.json


pankitgami:
Package use the `Illuminate\Console" to generate the test cases ([check here](https://github.com/laravel/dusk/blob/04a4ec585d6da7ef0ded12d269f8fec87beaad5b/src/Console/InstallCommand.php#L5)). But it is not listed in as a dependency in `composer.json` file ([check here](https://github.com/laravel/dusk/blob/master/composer.json)).

PR #232, opened 2017-04-19
[1.0.13] Add Illuminate/Console as dependency

deleugpn:
I was thinking about this yesterday and then #230 showed up to confirm my thoughts.


deleugpn:
Close #230

Issue #217, opened 2017-04-10
[Feature Request] Make text-based assertions case-insensitive


jkudish:
Hey there,

I was wondering if it would be possible to make text-based assertions, like `assertSeeIn` case insensitive?

A situation we've run into is that we'll uppercase something with css because of a design choice, and a test will start failing because that text is now no longer in the expected case, but the text is still the expected string, so it's not really a failure in our eyes.

Any thoughts?

browner12:
I would also like to see this, because I feel like having them case-sensitive makes the test unnecessarily brittle.  If you want I can make a PR for it?

deleugpn:
Go for it. Keep in mind that you might have to limit yourself to not make a breaking change (I make usage of sensitive assertion) while still offering the new assertion and Taylor might not accept it. It's an easy change nonetheless.

browner12:
yah, I already started. Everything will be BC.

browner12:
https://github.com/laravel/dusk/pull/218

jkudish:
@taylorotwell I kind of agree the solution in #218 locks us into a corner, but any other suggestions how we could resolve this?

browner12:
first off, i'm trying to rack my brain on other options we would put here, but I can't come up with any. any thoughts?

One option would be maybe an `$options` parameter.  Don't love this, but don't hate it.

Another option would be adding additional methods

- `assertSeeCaseInsensitive()`
- `assertSeeInCaseInsensitive()`
- `assertDontSeeCaseInsensitive()`
- `assertDontSeeInCaseInsensitive()`

but this adds a lot of LOCs, and the method names aren't particularly pleasing.

deleugpn:
All of this is why I suggested this as a custom method in your own extended Browser class. Dusk, as a whole, needs to attend to A LOT of use-cases and it would be unrealistic to expect that every assertion you need to make lives in MakesAssertion trait. That is why Browser is macroable and extendable. The most recent system I've been developing is reaching 100 dusk tests and 600 assertions, none of which requires sensitivity. I needed [Assert See Collection](https://github.com/laravel/dusk/pull/186), though, but I just added that in my extended browser and problem solved.

browner12:
yah, so what we're discussing is the general value of specific assertions.  If 99% of people would use it, it is something that should probably go in the core.  If it's only 50%, maybe that's something we suggest to make macroable.

IMHO, case insensitive text assertions are incredibly common, which is why I would suggest having it in the core.

[...]

Issue #209, opened 2017-03-31
Checking the assertion of form elements via a selector


votintsev:
At the moment assertionChecked (for example) is possible only through the name or id. But modern CSS styles often use an input[type=checkbox] only as a wrapper without a name & id. Currently I did something like this:

```
<?php

namespace Tests\Browser;

use Laravel\Dusk\Browser;
use Tests\DuskTestCase;
use PHPUnit_Framework_Assert as PHPUnit;

class SomeTest extends DuskTestCase
{
    public function testAddToFavorite()
    {
        $this->browse(function (Browser $browser) {
            $browser = $browser->visit('/some_page');
            $browser = $this->assertNotChecked($browser, '.block .add-to-fav input');
            $browser = $browser->click('.block .add-to-fav');
            $browser = $this->assertChecked($browser, '.block .add-to-fav input');
            $browser = $browser->visit('/some_page');
            $browser = $this->assertChecked($browser, '.block .add-to-fav input');
        });
    }

    public function assertChecked($browser, $selector)
    {
        $fullSelector = $browser->resolver->format($selector);
        $element = $browser->resolver->findOrFail($selector);

        PHPUnit::assertTrue(
            $element->isSelected(),
            "Expected checkbox [{$fullSelector}] to be checked, but it wasn't."
        );

        return $browser;
    }

    public function assertNotChecked($browser, $selector)
    {
        $fullSelector = $browser->resolver->format($selector);
        $element = $browser->resolver->findOrFail($selector);

        PHPUnit::assertFalse(
            $element->isSelected(),
            "Checkbox [{$fullSelector}] was unexpectedly checked."
        );

        return $browser;
    }
}

```

It work correctly, but I would like to have a more attractive solution. How realistic is it to expect the addition of an opportunity to use a selector to validate such assertion?

deleugpn:
You can just extend the Browser class and override the `newBrowser` method in your `DuskTestCase.php`. Place all your custom assertions in your extended Browser class and return a new object of it from `newBrowser`.

driesvints:
Feel free to create your custom assertion.

Issue #206, opened 2017-03-27
[Feature Request] Assert Message


mexcoder:
Hi, PHPUnit supports an extra argument in all its assertions methods to add a message regarding the error, like [this example](http://www.kreamer.org/phpunit-cookbook/1.0/assertions/display-useful-message-when-test-fails) and as is used on [Concerns/MakesAssertions.php](https://github.com/laravel/dusk/blob/master/src/Concerns/MakesAssertions.php#L252), could this parameter be added to the Dusk's assert methods so we can add our own messages?

deleugpn:
Unfortunately we have messages bound to the method itself. Example: 

```
	public function assertSeeIn($selector, $text)
	{
		$fullSelector = $this->resolver->format($selector);
		
		$element = $this->resolver->findOrFail($selector);
		
		PHPUnit::assertTrue(
			Str::contains($element->getText(), $text),
			"Did not see expected text [{$text}] within element [{$fullSelector}]."
		);
		
		return $this;
	}
```

This could not be replaced with a simple optional parameter `public function assertSeeIn($selector, $text, $message = 'Did not see expected text [{$text}] within element [{$fullSelector}].')` for obvious reasons.

I wouldn't recommend implementing this just for assertions that could simply add a new optional parameter because it could cause inconsistency and I also wouldn't recommend adding if statements through the whole trait just to offer this. 

If you really want something like this, I recommend you to extend the browser class and override all the MakeAssertions methods.

driesvints:
We'll probably not add this but the Browser class is macroable so you should be able to overwrite or add your own methods.

PR #234, opened 2017-04-19
[1.0.13] Include custom browser with scaffolding to encourage custom assertion

deleugpn:
We always want more assertion than the framework offers. This is a way of making it easy for everyone to understand how easy it is to build your custom assertion.

#217 #209 #206 #189 all talks about other options of assertion.


taylorotwell:
I don't think I will include this out of the box. Let's just have docs on how to use Browser macros.

browner12:
do we have docs on this @deleugpn ?

deleugpn:
I didn't write any docs at the time of this because i didn't know how to use macros then. Now I understand what Taylor wanted. Macros rock

browner12:
yah, i think I got 1 working for an assertion, now just trying to figure out the best place to register it. maybe we should add a suggestion to the docs?

PR #236, opened 2017-04-19
[1.0.13] Better approach for environment exception

deleugpn:
Extract environment to class attribute. If somebody wants to override the class and explicitly register Dusk on production, it won't juts be an accident.

Also added tests for this.

### DIFF
--- a/src/DuskServiceProvider.php
+++ b/src/DuskServiceProvider.php
@@ -8,6 +8,14 @@ use Illuminate\Support\ServiceProvider;
 
 class DuskServiceProvider extends ServiceProvider
 {
+
+    /**
+     * Environments prohibited to register dusk
+     *
+     * @var array
+     */
+    protected $blacklist = ['production'];
+
     /**
      * Bootstrap any package services.
      *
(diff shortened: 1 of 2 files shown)


browner12:
would it be better to switch to a whitelist?

that way someone doesn't accidentally delete the entry or misspell it, and having it working on production.

deleugpn:
@browner12 I'm open to suggestion on how to whitelist every possible word people in the world might be using. Even if we go as far as making a huge whitelist, it still might not be enough. How would we start?

```
local
development
develop
testing
stage
staging
test
```

browner12:
yah, I see what you're saying.

if we wanted to whitelist, we would definitely have to implement a publishable `config/dusk.php` for people to add their options to.

taylorotwell:
I dunno, I don't like extending a service provider just to change this.

taylorotwell:
I'm not sure I like a config file either.

deleugpn:
@taylorotwell if someone explicitly wants to shut off the production exception, they're going to have to extend the Service Provider or edit code direct in vendor. This was simply to make it easier in case there's a genuine reason to want to register Dusk on production.

PR #51, opened 2017-01-24
Use consistent php binary for running dusk tests

splatEric:
When trying to get dusk tests running on codeship, I ran into a problem where the php that was being triggered from the Process was call not consistent with the php binary I was using to run the command. This pull request tackles this by making using of the ```PHP_BINARY``` constant to use the same version of php.

I wrote this defensively as I wasn't sure if there might be circumstances under which the ```PHP_BINARY``` constant was not set.

PR #240, opened 2017-04-20
Adds back PHP_BINARY to phpunit process

jhoff:
Disclaimer: Not sure of the implications of this on windows, as I can't test it...

A while back a [pull request](https://github.com/laravel/dusk/pull/51/files) was accepted that ensured PHP_BINARY was used for phpunit. Almost immediately after, that change was reverted in a refactor. This brings that fix back.

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -78,11 +78,11 @@ class DuskCommand extends Command
     /**
      * Get the PHP binary to execute.
      *
-     * @return string
+     * @return string|array
      */
     protected function binary()
     {
-        return PHP_OS === 'WINNT' ? base_path('vendor\bin\phpunit.bat') : 'vendor/bin/phpunit';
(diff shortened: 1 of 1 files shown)


taylorotwell:
So why are we adding it back?

jhoff:
This change simply makes sure that phpunit is run with the same version of php that you executed artisan with.

In our case, we are running tests with different versions of php depending on the project we're currently working with so we can't rely on $PATH to resolve the correct version for us.

taylorotwell:
So why was it reverted?

jhoff:
As far as I can tell, the original fix was accepted and then in the very next merge commit it was gone completely:

image

https://github.com/laravel/dusk/commit/db89a685060d746486a5f37d59afb7addada7ad7
https://github.com/laravel/dusk/commit/da7508e4a13cbf5e1efd448573b57a3a62d0ee58

If you trace back the commit parents on that second merge commit, it looks like the refactor was here:

https://github.com/laravel/dusk/commit/00338d5afa5892c45ff01af62a110855c8ea2771

PR #243, opened 2017-04-22
Revert "Adds back PHP_BINARY to phpunit process"

deleugpn:
Reverts laravel/dusk#240

This breaks Dusk on Virtual Machines with Windows Hosts. After updating to `1.0.13`, this is the result: 

```
[dusk@php7 linus-laravel]$ php artisan dusk --group=f

dir=$(d=${0%[/\\]*}; cd "$d"; cd '../phpunit/phpunit' && pwd)

# See if we are running in Cygwin by checking for cygpath program
if command -v 'cygpath' >/dev/null 2>&1; then
        # Cygwin paths start with /cygdrive/ which will break windows PHP,
        # so we need to translate the dir path to windows format. However
        # we could be using cygwin PHP which does not require this, so we
        # test if the path to PHP starts with /cygdrive/ rather than /usr/bin
        if [[ $(which php) == /cygdrive/* ]]; then
                dir=$(cygpath -m "$dir");
        fi
fi

dir=$(echo $dir | sed 's/ /\ /g')
"${dir}/phpunit" "$@"
[dusk@php7 linus-laravel]$
```

By erasing the `PHP_BINARY` from the array, I was able to run my tests normally again.

@jhoff any reason why you need this change?

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -78,11 +78,11 @@ class DuskCommand extends Command
     /**
      * Get the PHP binary to execute.
      *
-     * @return string|array
+     * @return string
      */
     protected function binary()
     {
-        return PHP_OS === 'WINNT' ? base_path('vendor\bin\phpunit.bat') : [PHP_BINARY, 'vendor/bin/phpunit'];
+        return PHP_OS === 'WINNT' ? base_path('vendor\bin\phpunit.bat') : 'vendor/bin/phpunit';
     }
 
     /**


jhoff:
Interesting. It seems that Cygwin needs to depend on $PATH to resolve php. What is the value of PHP_BINARY in your environment?

We needed it for the reasons stated in the recent (merge request)[https://github.com/laravel/dusk/pull/240#issuecomment-295805055]. When running the `dusk` command from artisan, using PHP_BINARY will ensure that the process runs with the same version of php that you executed artisan with. Without it, then you have to do a bunch of $PATH mangling to get stuff to work, especially if you work with multiple versions of php on a system.

deleugpn:
@jhoff  `PHP_BINARY` resolves to `/usr/bin/php`.

taylorotwell:
OK I'll make a note so I don't get duped by this again. Brutal.

jhoff:
Sorry for this being an issue. We'll have to work through it on our end then. If I had a windows machine to test this out I might have found the issue.

deleugpn:
@jhoff that's what the community is for :)
I'm still curious as to why Symphony process is dumping the script on the screen. I'm going to see if I can dig it further sometime.

deleugpn:
@jhoff I'm really interested in knowing how THAT is a solution for you. Follow my thoughts: 

#### [Symphony will build the command line that needs to be executed](https://github.com/symfony/process/blob/3.2/ProcessBuilder.php#L267-L268)

At this point, having `PHP_BINARY` set will give you something along the lines of 
```
"'/usr/bin/php' 'vendor/bin/phpunit' '-c' '/var/www/html/dusk-project/phpunit.dusk.xml'"
```
Which basically means: PHP Binary, please run this **bash script** (vendor/bin/phpunit) with these arguments. But PHP doesn't know how to process a bash script, so it just outputs it's content.

From what you've told us, by running it like this, your PHP Binary is capable of running the `phpunit` bash script in such a way that PHPUnit will be executed by the same version it was initially invoked. How is this possible?

PR #247, opened 2017-04-25
Assure the environment running dusk matches the web server environment

deleugpn:
Dusk tries to enforce environment compatibility between itself and the environment the browser will access by swapping the `.env` file, which will be used by the Web Server. Doing this allows cool things like `assertDatabaseHas` to work properly. However, if you don't have compatible setup, things might not work correctly.

One way to get stuck is to combine `dusk` with `serve`, because `php artisan serve` will use `.env` before dusk had a chance to swap it. This PR makes it easier for the developer not to mess with the local environment by running dusk on it as well as making it easy to identify which environment dusk was running and which environment the webserver was providing.

### DIFF
--- a/src/DuskServiceProvider.php
+++ b/src/DuskServiceProvider.php
@@ -29,6 +29,13 @@ class DuskServiceProvider extends ServiceProvider
             'middleware' => 'web',
             'uses' => 'Laravel\Dusk\Http\Controllers\UserController@user',
         ]);
+
+        Route::get('/_dusk/env', [
+            'middleware' => 'web',
+            'uses' => function () {
+                return getenv('APP_ENV');
+            }
+        ]);
     }
 
     /**
--- a/src/TestCase.php
+++ b/src/TestCase.php
@@ -122,6 +122,8 @@ abstract class TestCase extends FoundationTestCase
             static::$browsers->push($this->newBrowser($this->createWebDriver()));
         }
 
+        $this->validateBrowserEnvironment(static::$browsers->first());
+
         return static::$browsers;
     }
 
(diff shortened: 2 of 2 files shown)


taylorotwell:
assertDatabaseHas already works properly though?

deleugpn:
It does if you're on the right environment. Suppose the following: 

```
php artisan serve // This will be .env local file
php artisan dusk  // This will go to .env.dusk.local
```

Dusk will run your tests against DB_DATABASE set on `.env.dusk.local`, but your browser will make changes to the database set on `.env`. `assertDatabaseHas` will obviously fail. 

**My point was** that dusk needs to ensure compatibility between environments to properly run. This PR adds an extra layer to ensure that as well as help developers identify a bad setup.

SebastianS90:
`serve` should already look up the `.env` file for each single request and see changes that are made by dusk.
Exception: If you use `config:cache` then it won't use the `.env` file anymore. That is why you should not use `config:cache` for development.

I quite like the possibility to have different environments for the webserver and dusk: Imagine you deploy your app to the staging environment, i.e. a webserver running somewhere with a setup similar to your production setup. Then you can start dusk on your local machine and run your test suite against the staging deployment. Obviously you won't be able to use the quick login route if dusk is disabled in your staging environment, but you still could login the user by filling the normal login form.

deleugpn:
@SebastianS90 your use-case makes total sense, although you lose some functionalities like `assertDatabaseHas`. But your statement about `serve` sync with changes never happened to me.

SebastianS90:
Sorry, it actually looks like the internal webserver doesn't reload `.env` for subsequent requests, so I got that part wrong.

Instead of throwing an Exception you could add a one-line warning to the dusk console to make the developer aware of this fact. That would allow for the use-case described above (also make sure that the tests don't fail only because of the newly introduced route is missing).

But the better solution might be to clearly document how you can let dusk use a separate `.env.dusk` file together with the internal php webserver. I guess that Taylor uses Laravel Valet (which uses a normal webserver and therefore reloads `.env` on every single request). But that is a Mac-only solution.
For having it work together with the internal php webserver we actually might need something like the proposed changes from #249.

deleugpn:
@SebastianS90 that is definitely a great a idea. However, it would somewhat affect the code structure because `TestCase` doesn't know the existence of `$this->output` located in the Service Provider. The Service Provider cannot check the environment because it needs the browser to be instantiated. I'm going to try and look over the code and see if I can find a clean way to just display the warning.

Issue #162, opened 2017-03-01
Dusk env file not working with artisan serve


renedekat:
When running php artisan serve and then in another terminal php artisan dusk, the env file isn't loaded properly.

I think Dusk should launch php artisan serve after copying the Dusk specific .env file, and then kill the process afterwards.

SebastianS90:
Make sure that you do not cache the configuration, i.e. run `php artisan config:clear` before starting dusk. Otherwise you will have cached configuration based on your normal `.env` file and `.env.dusk` will be ignored.

deleugpn:
I like the thought of having dusk serving the application automatically instead of having to do this manually. Perhaps something to be done at DuskTestCase?

SebastianS90:
@deleugpn you mean that you do not wan't to run `php artisan serve` manually?
Dusk expects you to have a running webserver at `APP_URL`, this can be:
* a `php artisan serve` development server
* Laravel Valet serving your app on `app.dev`
* Your staging environment using nginx with a configuration similar to your production environment. This allows for browser tests to run against your real-world environment and you might find bugs that are not present on your development machine. Pretty cool!

The only situation where it might actually be useful to have a webserver booted up by Dusk itself is if that will make mocks usable in Dusk test cases. But that is not possible so far. See also the discussion in #152.

deleugpn:
Yes, that's exactly what I mean. I work on centos. Since Google Chromw doesn't support Red Hat, I installed a Ubuntu server headless just to run the tests. My activities in this machine is:
- turn on; 
- run tests; 
- realize you forgot to run serve; 
- run serve; 
- run tests; 

Since my centos runs on another VM, I thought to be over complicated to have Ubuntu connecting there for the sake of dusk. Serve would be enough. Given this idea, today I'm going to try and have DuskTestCase turn on the server for me.

SebastianS90:
@deleugpn instead of having `DuskTestCase` start the webserver for every single request you could also let `DuskCommand` do the job once for your whole test suite:

Create a class in `app/Console/Commands`:
```php
<?php

namespace App\Console\Commands;

use Laravel\Dusk\Console\DuskCommand as BaseCommand;
use Symfony\Component\Process\ProcessBuilder;

class DuskCommand extends BaseCommand
{
    public function handle()
    {
        $webserver = (new ProcessBuilder())
            ->setTimeout(null)
            ->add('exec')
            ->add(PHP_BINARY)
            ->add('artisan')
            ->add('serve')
[...]

PR #249, opened 2017-04-25
Add dusk:serve command

deleugpn:
While keeping `php artisan dusk` unaltered, this PR will add `php artisan dusk:serve` to the list of available commands to achieve an automatic behavior of serving the application.

- Swap `.env` files;
- Build and run `php artisan serve`
- Build and run Phpunit (inheriting `DuskCommand`)

Not only solves the tedious task of having to turn on a server to run dusk everytime, it will never `serve` an `env` that doesn't match dusk's because of the order of processing.

Special thanks to @SebastianS90 for the `php artisan serve` builder posted on #162 

It even works on `cmd.exe` (which I don't really find important, but hey).

cmd.exe

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -41,6 +41,10 @@ class DuskCommand extends Command
         parent::__construct();
 
         $this->ignoreValidationErrors();
+
+        $this->purgeScreenshots();
+
+        $this->purgeConsoleLogs();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't plan on adding this to the core. The built-in PHP web server has some serious limitations and I don't want to always be answering questions related to that. Thanks though.

deleugpn:
@taylorotwell can I resubmit just with the refactoring of `DuskCommand` to be able to use custom command?

SebastianS90:
It would be nice to have the `withDuskEnvironment` method as a trait, then it can be easily reused without inheriting everything form `DuskCommand`.

deleugpn:
@SebastianS90 problem with extracting withDuskEnvironment to a trait is that it cannot be called twice in a single process, otherwise you end up overriding the `.env.backup` and might even end up with a third environment set.

PR #250, opened 2017-04-25
Refactor to allow extending dusk's behavior

deleugpn:
Makes it possible to be able to extend `DuskCommand` and call `runPhpunit()` after custom preparation.


taylorotwell:
I don't like performing actions in constructors.

PR #251, opened 2017-04-26
Add missing assert messages

kszcode:
Sometimes it's nice to write an assert message on why did a certain
path failed to be reached.

PS: also moved the protected function `assertHasQueryStringParameter` after the public functions. 
I was automatically done by PHP Storm code formatter.
(perhaps I should have split the PR's)


taylorotwell:
No plans to add all these.

PR #254, opened 2017-04-27
Added new interaction to type without clearing

alexgarrett:
In certain situations, appending onto already typed values is useful, particularly when continuing to retry inline validation and assert the result.

Or, just a more elegant way to type after line breaks have been added using `keys`.

```php
->type('@body', 'Test message')
->keys('@body', '{shift}', '{enter}')
->append('@body', 'New line')
```

### DIFF
--- a/src/Concerns/InteractsWithElements.php
+++ b/src/Concerns/InteractsWithElements.php
@@ -178,7 +178,7 @@ trait InteractsWithElements
      * @param  string  $value
      * @return $this
      */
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why not just call it `append`?

alexgarrett:
Yeah makes sense the more I think about it. Updated.

taylorotwell:
Thanks

PR #260, opened 2017-05-05
Add a position method to set the position of the browser window

calebporzio:
Given the size of my screen, I like to load the browser windows more in the middle of the screen.

I added a position method on `Browser.php` similar to `resize`:
```
$browser->position(150, 300);
```

Thanks!

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -5,6 +5,7 @@ namespace Laravel\Dusk;
 use Closure;
 use BadMethodCallException;
 use Illuminate\Support\Str;
+use Facebook\WebDriver\WebDriverPoint;
 use Illuminate\Support\Traits\Macroable;
 use Facebook\WebDriver\WebDriverDimension;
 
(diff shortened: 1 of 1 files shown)


deleugpn:
Just in case this doesn't get merged, you know that you can add it to your own Browser class by overriding the `newBrowser` method on `DuskTestCase.php`, right?

taylorotwell:
I think this is maybe something you could add to newBrowser as suggested so it applies easily to every test.

calebporzio:
@taylorotwell Ya, I actually did it there, I just wanted a nicer interface (instead of having to write: `->setPosition(new WebDriverPoint($x, $y));`

```
return (new Browser($driver))->resize(700, 500)->position(350, 300);
```

not the end of the world though.

deleugpn:
I don't see any reason you wouldn't be able to achieve the same API using `DuskTestCase`

calebporzio:
Maybe I'm misunderstanding. I certainly could do that, I'd just rather have it available by default and for others, thats why I PRed it. Or am I missing something @deleugpn?

deleugpn:
I understand the motive behind the PR. What I don't understand is when you say that not having available by default makes you have to call it in a different way.

calebporzio:
Sorry, I really think theres some gap in understanding here. So here's my understanding of your suggestion:

1. I override the `newBrowser` method in my projects DuskTestCase (the stub)
2. I do one of the following:
 - return a customized `Browser` class that has my `position` method on it
 - directly set the position in the `newBrowser` method using `setPosition`

What I mean by "default" is "backed into this repo so I don't have to maintain a custom `Browser` file and others can take advantage of the cleaner syntax of `position` instead of `driver->setPosition`

Let me know where I'm off @deleugpn thanks!

PR #261, opened 2017-05-05
Add waitForReload method to wait for page reload to continue

calebporzio:
When writing Laravel & Vue.js I often call an endpoint in my app:
```
axios.get('/endpoint').then(response => {
    window.location.reload(); //or
    window.location.href = response.data.redirect_url;
});
```
This can be difficult to test because of reload timing.

The following is my solution:
```
$browser->waitForReload(function ($browser) {
    $browser->click('.some-action');
})->assertSee('something');

// or inline works too

$browser->click('.some-action')
    ->waitForReload()
    ->assertSee('something');
```

The implementation uses Javascript to set a variable and check that it no longer exists (after reload). It works for me, but I can see that some would feel iffy about this solution. Hopefully it lands well here and is pulled in.

Thanks for all the hard work as always!

### DIFF
--- a/src/Concerns/WaitsForElements.php
+++ b/src/Concerns/WaitsForElements.php
@@ -121,6 +121,26 @@ trait WaitsForElements
     }
 
     /**
+     * Wait for the current page to reload.
+     *
+     * @param  Closure  $callback
+     * @param  int  $seconds
+     * @return $this
+     */
+    public function waitForReload($callback = null, $seconds = 5)
+    {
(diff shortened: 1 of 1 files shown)


taylorotwell:
`$token` was not imported into  the Closure properly. I fixed that but you may want to make sure this works.

calebporzio:
Shoot, thanks. I did it right in my local implementation.

It works.

Thanks!

PR #266, opened 2017-05-10
Add tinker method to pause and open Laravel Tinker

calebporzio:
I recently stumbled on a cool trick that has made writing Dusk tests SOOO much nicer for me.

Take a look at the following code:
```
$browser->visitRoute('route');

// special snippet that pauses execution and opens Tinker
eval(\Psy\sh());

@browser->assertSee(...
```
You can see it in action here:

I also wrote about [here](https://blog.tighten.co/supercharge-your-laravel-tinker-workflow)

I would love to have this functionality inline like so:
```
$browser->visitRoute('route')
    ->tinker()
    ->assertSee(...
```

I'm guessing there are some problems with my implementation, like it wouldn't work if someone used Dusk outside the Laravel App context, or if someone chose to not include the Laravel tinker package. Let me know if you have any thoughts on this and we can work towards a better solution, unless what I have is good enough!

As always, thanks for all the hard work! Dusk is AWESOME :)

- Caleb

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -322,6 +322,18 @@ class Browser
     }
 
     /**
+     * Pause execution of test and open Laravel Tinker (PsySH) REPL.
+     *
+     * @return $this
+     */
+    public function tinker()
+    {
+        eval(\Psy\sh());
+
+        return $this;
+    }
+
+    /**
      * Stop running tests but leave the browser open.
      *
      * @return void


taylorotwell:
Nice thanks.

sileence:
Wow, this is very nice! Thank you <3

bobthecow:
With this change, dusk should prolly add a dependency on `laravel/tinker` (or recommend it, and do a runtime check when calling `tinker` and error if psysh isn't available).

nathanjisaac:
@calebporzio my chrome windows keeps closing automatically after I run `eval(\Psy\sh());`.  Any tips?

calebporzio:
Hmm, odd - never ran into that. Maybe share more specifics?

nathanjisaac:
Sure thing. I'm using the following. 

Windows 10
PHP: 7.0.16
Chrome: Version 59.0.3071.115
Laravel: 5.4.25
Dusk: 1.1.0

I have been doing some digging. Even this will close the browser window. From the PHP docs on that method it should keep the window open.

```
$browser->visitRoute('route');
$browser->stop();
```

This does exit phpunit and that seems to be when the window closes. 

Still investigating...

arya48:
is this still working in Laravel 8 ? 

It seems that the line `eval(\Psy\sh());` is being skipped in my code after upgrading to Laravel 8.

Issue #100, opened 2017-02-03
Test methods influence each other through session


fschwaiger:
Test methods should not influence each other. However, the following two methods share the same session and the second will fail, because the user is already logged in.

```php
    public function test_login_redirects_on_first_time()
    {
        $this->browse(function (Browser $browser) {
            $browser->visit('/') // has RedirectIfAuthenticated middleware to /dashboard
                ->type('username', 'johndoe')
                ->type('password', 'secret')
                ->press('@login')
                ->assertPathIs('/dashboard');
        });
    }

    public function test_session_from_previous_method_should_not_exist()
    {
        $this->browse(function (Browser $browser) {
            $browser->visit('/')
                ->assertPathIs('/');  // actually is /dashboard
        });
    }
```

The issue is fixed if the last line, `closeAllButPrimary()` is changed to `closeAll()`:

```diff
    public function browse(Closure $callback)
    {
        $browsers = $this->createBrowsersFor($callback);

        try {
            $callback(...$browsers->all());
        } catch (Exception $e) {
            $this->captureFailuresFor($browsers);

            throw $e;
        } catch (Throwable $e) {
            $this->captureFailuresFor($browsers);

            throw $e;
        } finally {
-            static::$browsers = $this->closeAllButPrimary($browsers);
+            static::$browsers = $this->closeAll($browsers);
        }
    }
```

Sieabah:
You should drop the session each test instead of killing the entire chromedriver process. Plus, why aren't you mocking the session in your tests?

fschwaiger:
Mocking the session is not possible since the browser is a different process.
Ensuring the test environment is clean should be the job of `browse(...)`, `setUp()` or `tearDown()`.

JVMartin:
This is an issue for me as well.

And it is not possible to use "array" for the session.

Is there a workaround for now?

georaldc:
@Sieabah You can't use array because as already mentioned, the actual tests execute under a different process. Probably the same reason you cannot use in-memory sqlite. One way to circumvent these problems is to use a physical sqlite file (if you want to use sqlite), set both cache and session drivers to use the database and use the DatabaseMigrations trait to continually reset your database before every test. Of course, this means you would need the migration files for both cache and session tables, even if you will be using something else on a different environment like production.

Because of this though (running on a separate process), I just realized that this prevents me from setting up my application environment such as mocking services and disabling middleware since anything that happens in my DuskTestCase classes won't exist during the actual test runs. Any way around this?

[...]

PR #268, opened 2017-05-12
Reset cookies after browse closure completes

inxilpro:
Right now, each call to `browse()` shares the same `Browser` instance, which means that tests that open sessions or set other cookies can influence each other. This adds a `clearCookies()` method the `Browser` class, and automatically calls it after the `browse()` closure completes. That way each test starts fresh. This should resolve issue #100.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -344,6 +344,16 @@ class Browser
     }
 
     /**
+     * Clear all browser cookies.
+     *
+     * @return void
+     */
+    public function clearCookies()
+    {
+        $this->driver->manage()->deleteAllCookies();
+    }
+
+    /**
      * Dynamically call a method on the browser.
      *
      * @param  string  $method
--- a/src/TestCase.php
+++ b/src/TestCase.php
@@ -100,7 +100,7 @@ abstract class TestCase extends FoundationTestCase
         } finally {
             $this->storeConsoleLogsFor($browsers);
 
-            static::$browsers = $this->closeAllButPrimary($browsers);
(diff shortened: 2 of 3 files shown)


taylorotwell:
That's the whole point of it staying open I believe. If you don't want that behavior you can start a fresh browser for each test.

inxilpro:
I don't think it's an obvious behavior or one that's consistent with typical testing where the expectation is that tests run in isolation. How would you feel about exposing it as an option on the Dusk test case? Looking at the issue I referenced, it doesn't look like I'm alone.

ryancwalsh:
I have spent hours working on this. I'd love to know how to have just one test class that has 3 test cases... where Test 2 has a new browser started (separate from Test 1, so all cookies are cleared, etc), but then Test 3 continues using the same browser as Test 2. Is that possible? How?

Thanks @inxilpro and @taylorotwell . I have found your various Github conversations helpful (not to mention all of Laravel, which is mind-blowingly awesome).

ryancwalsh:
I think this is working for me (a new utility method within my test class):

    protected function clearCookies() {
        foreach (static::$browsers as $b) {
            Log::debug('browser=' . json_encode($b));
            Log::debug('cookies=' . json_encode($b->driver->manage()->getCookies()));
            $b->driver->manage()->deleteAllCookies(); //https://github.com/laravel/dusk/pull/285/files
            Log::debug('cookies=' . json_encode($b->driver->manage()->getCookies()));
        }
    }

I still don't understand Webdriver and how there are multiple "browsers", but I think I'm making progress.

PR #285, opened 2017-05-31
Optionally clear cookies during tearDown()

inxilpro:
_This is an alternate version of #268_

Right now, an entire test class shares the same primary `Browser` instance (secondary instances are destroyed after a test, but the primary instance is only cleared after the entire test class has finished running). This means that separate tests that open sessions or set other cookies can influence each other.

This PR adds a `clearCookies()` method the `Browser` class, and adds a `$clearCookiesBetweenTests` property (defaulted to `false`) to `Laravel\Dusk\TestCase` that lets you configure dusk to clear cookies between tests.

This means that if you want your tests to run in a more isolated way, you can just set:

```php
class ExampleTest extends DuskTestCase
{
    protected $clearCookiesBetweenTests = true;

    // ...
}
```

This should resolve issue #100.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -344,6 +344,16 @@ class Browser
     }
 
     /**
+     * Clear all browser cookies.
+     *
+     * @return void
+     */
+    public function clearCookies()
+    {
+        $this->driver->manage()->deleteAllCookies();
+    }
+
+    /**
      * Dynamically call a method on the browser.
      *
      * @param  string  $method
--- a/src/TestCase.php
+++ b/src/TestCase.php
@@ -57,7 +57,7 @@ abstract class TestCase extends FoundationTestCase
     }
 
     /**
-     * Optionally clear all cookies after a test has run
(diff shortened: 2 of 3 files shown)


taylorotwell:
I think I will just let people add this to their own tear down for now.

HSPDev:
We just hit this today. It was actually expected behaviour for us to have each test completely seperate. Otherwise you are relying on the order of the tests being run.

We have an account switching feature in our SaaS platform, and it proved to be a little buggy with its remains in the session. We had it higher up in our testcase, but a function being run after 5-6 other tests "blew up" because of that. It took ~2 hours to track the bug down because we thought the cookies got cleared and researched everything else; At least we got to know Laravel better. :grin: 

Please consider reopening this.

dsandber:
I think most users assume regression will be independent of each other.  Not clearing the cookies violates this.

Please re-consider making the cookies get cleared by default, or if not, at least add an option to do it.

Thanks!

PR #287, opened 2017-06-04
Click using a jQuery selector

nicolasbeauvais:
Similar to the `clickLink` method but to click any kind of element using a more complex selector.

Allow the use of things like: `input[name=email]:first + span`


taylorotwell:
I would rather not have two methods for clicking.

PR #290, opened 2017-06-08
support auto-discovery coming in Laravel 5.5

browner12:


deleugpn:
This would cause problems because Dusk would be registered in production.

browner12:
that's why you only put dusk in your `require-dev`.

now you don't need conditional service provider registering.

taylorotwell:
Went ahead and branched off 1.0 branch so we can keep master branch as our future 2.0.

PR #291, opened 2017-06-08
Support auto-discovery

m1guelpf:
Support package autodiscovery coming in Laravel 5.5
This also removes the registering-only-in-local step, as it can be only required on require-dev


browner12:
beat you to it by a couple hours 😉 

https://github.com/laravel/dusk/pull/290

m1guelpf:
@browner12 Good timing!

PR #293, opened 2017-06-14
Add message parameter to all MakesAssertions trait assertion methods

damiantw:
This pull request allows for specifying custom PHPUnit messages for all assertion methods within the Dusk MakesAssertions trait. There are no breaking changes and existing messages will be retained if no message is specified.


taylorotwell:
No plans to add this to all the signatures right now in case I want to add other parameters to these methods.

PR #299, opened 2017-06-25
Preview mid test

assertchris:
This change adds the ability to preview the screen state mid-test. It obviously needs docs still, but I thought the code would illustrate it well enough to decide whether the additional polish was warranted...

Edit 1: it only works on macOS. I'm sure it should be possible to cater for windows and linux (Ubuntu at least) but I don't have those on hand to write the code for...

Edit 2: This is obviously less useful if you're not running headless.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -338,6 +338,21 @@ class Browser
         return $this;
     }
 
+    public function preview($interval = 1000)
+    {
+        $time = time();
+
+        $name = sprintf('%s/%s.png', rtrim(static::$storeScreenshotsAt, '/'), $time);
+        $command = sprintf('/usr/bin/qlmanage -p %s > /dev/null 2> /dev/null', $name);
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Cool. What does this do that is different than sticking something like a 1000ms pause in the middle of your test?

assertchris:
In headless mode, it's a way to answer the question; "why is this next assertion failing".

taylorotwell:
Ah, gotcha.

taylorotwell:
Is there any way this could just be worked into the normal screenshot function? It feels a *little* strange to me to have a different function for taking a screenshot if your headless. To me, if I'm the user, I shouldn't have to be concerned about that detail? Thoughts?

taylorotwell:
Closing for now. Would prefer to work it into screenshot() method if at all possible. I mentioned this to @themsaid.

PR #314, opened 2017-07-10
Added assertSeeButton() and assertDontSeeButton()

Janiaje:
Original methods could assert if a button is visible or not.


taylorotwell:
You'll need to import that exception.

Janiaje:
Sorry, I imported it, when a tested it, but forgot to change it here.
Imported it in https://github.com/laravel/dusk/pull/314/commits/1a6f2187f68a7335a57b3c0422ae3b7da336d42a

PR #319, opened 2017-07-19
Upgrading to chromedriver 2.30 + Default to chrome headless

dbpolito:
I see we have a good example of this on docs, but why isn't this the default configuration?

https://laravel.com/docs/5.4/dusk#chromedriver-options

This probably fix all homestead/docker related errors as it basically works without UI... also simplifies TravisCI and CircleCI integration.

At least on my tests with Docker, which really simplifies the setup, it's also 2x faster.

### DIFF
--- a/stubs/DuskTestCase.stub
+++ b/stubs/DuskTestCase.stub
@@ -3,6 +3,7 @@
 namespace Tests;
 
 use Laravel\Dusk\TestCase as BaseTestCase;
+use Facebook\WebDriver\Chrome\ChromeOptions;
 use Facebook\WebDriver\Remote\RemoteWebDriver;
 use Facebook\WebDriver\Remote\DesiredCapabilities;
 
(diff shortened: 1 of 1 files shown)


stubs/DuskTestCase.stub
```diff
@@ -28,8 +29,19 @@ abstract class DuskTestCase extends BaseTestCase
      */
     protected function driver()
     {
+        $options = (new ChromeOptions)->addArguments([
+            '--headless',
+            '--disable-gpu',
```
dbpolito: this is here as there is a recommendation to add this for now here: https://developers.google.com/web/updates/2017/04/headless-chrome

taylorotwell:
Can you tweak the Dusk test case stub on the "master" branch instead. That is where I've already added headless support. But, can you make sure the binaries are up to date and add the disable-gpu flag. I'll get it merged in when sent. Thanks.

dbpolito:
I swear i searched where you put it, but i missed it... :(

Anyway, looks like the bins are up to date and you were faster than me and added it there... thank you.

I added a comment on the commit you made: https://github.com/laravel/dusk/commit/115247c8d7015e82a432dbdefa9dc9cd7609880c#commitcomment-23180036

PR #339, opened 2017-08-21
Rename '$browser->with()' to '$browser->within()'

calebporzio:
```php
// current method
$browser->with('selector', ...

// more appropriate IMO
$browser->within('select', ...
```

Personally "within" is self explanatory. "with" doesn't seem intuitive to me and I believe the same for new-comers.

Also, I've been using "Pages" for reusable components and parts of a page. I'd eventially like to pull request the concept of "Components" or modify the way "Pages" behave to accommodate this. Using Page objects for UI flows and re-usable page components is common: see Ben Orenstiens video [here](https://www.youtube.com/watch?v=u6grFClqt6E&feature=youtu.be&__s=3zqnnsygjob5sobbuxzm)

```php
$browser->on(new Homepage)
    ->within(new Navbar, function ($browser) {
        $browser->click('@profile-dropdown'...
...
``` 

I understand this is a pretty big change to Dusk's API, but I think it's worth it in the long run for being added to 2.0.

Thanks!
Caleb

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -248,7 +248,7 @@ class Browser
      * @param  \Closure  $callback
      * @return $this
      */
-    public function with($selector, Closure $callback)
+    public function within($selector, Closure $callback)
     {
         $browser = new static(
             $this->driver, new ElementResolver($this->driver, $this->resolver->format($selector))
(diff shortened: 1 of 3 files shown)


taylorotwell:
Would rather not really make the breaking change. Just add ->within() and make it call ->with :)

deleugpn:
Perhaps try to make `within` and keep `with` calling `within` and marked as deprecated.

PR #346, opened 2017-08-29
Add assertHaveSelector function

amree:
Adding a selector where you can specify the amount of elements available. Example use case:

* Click a button that add rows to table
* Assert table has 2 rows


taylorotwell:
I think I will let this be macro'd in for now.

PR #347, opened 2017-08-29
Add @dusk blade directive to render dusk hooks in markup

calebporzio:
This PR adds a blade directive to render `dusk=""` attribute "hooks" in markup:

```
<div @dusk('foo')></div>
<div dusk="foo"></div> // compiles to
```

Note: because `DuskServiceProvider` isn't loaded in production, these directives will not be rendered in production, therefore leaving behind `@dusk('')` in rendered html. I will submit a PR to `laravel/framework` that compiles an empty string when `@dusk` is encountered in production.

### DIFF
--- a/src/DuskServiceProvider.php
+++ b/src/DuskServiceProvider.php
@@ -29,6 +29,12 @@ class DuskServiceProvider extends ServiceProvider
             'middleware' => 'web',
             'uses' => 'Laravel\Dusk\Http\Controllers\UserController@user',
         ]);
+
+        Blade::directive('dusk', function ($expression) {
+            $selector = trim(preg_replace("/[\(\)\\\"\']/", '', $expression));
+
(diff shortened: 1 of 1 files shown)


deleugpn:
Have you tested if this overrides an existing directive on the framework?

calebporzio:
@deleugpn - yes

taylorotwell:
If we're going to have a dusk directive in the framework anyways you might as well just put it there haha :)

calebporzio:
Ugh @taylorotwell, sorry that this is still a thing.

I actually think it's better to have both. This way we don't have to guess which environments the users want the `dusk=""` attributes rendered in (`local`, `testing`). They only load when `DuskServiceProvider` loads. It also makes the code much simpler because there's no expression parsing for environments and conditionals for rendering.

Let me know though and I'll just re-PR the other `laravel/framework` only implementation.

Issue #341, opened 2017-08-22
Only first test finshes, chromedriver doesn't stop


rostockahoi:
`chromedriver-linux` spawns when running `artisan dusk` but it doesn't stop after test execution, so every subsequent test will never finish because there cannot be multiple instances of chromedriver on the same port.

When I try to exec chromedriver manually it says:
`[0,002][SEVERE]: bind() returned an error, errno=98: Die Adresse wird bereits verwendet` which means "the address is already in use".

However if I kill `chromedriver-linux` manually it works again - once.

How can I fix the missing stop instruction?

calebporzio:
@rostockahoi - in your `TestCase.php` make sure your `prepare` method has the `@beforeClass` annotation like so:

```php
/**
     * Prepare for Dusk test execution.
     *
     * @beforeClass
     * @return void
     */
    public static function prepare()
    {
        static::startChromeDriver();
    }
```

Let me know if that doesn't fix your issue.

rostockahoi:
Hi @calebporzio, thanks for the reply. The line exists. That's why the first test run works. But as soon as there's a process chromderiver-linux that is not terminated after the test has finished the next chromedriver-linux instance runs into an error.

Something regarding the chromedriver-linux termination seems to not work properly.

Btw: I didn't change Dusk's code.

rostockahoi:
I think I've come closer to the problem. It seems everything is working just fine in `vendor/laravel/dusk/src/SupportsChrome.php`. It starts and stops the `chromedriver-linux` process as expected. However, on my machine (Linux Mint 18.2) there are spawning 2 processes of which only one gets terminated after the test has finished. The other one remains running and is then blocking port `9515`. The blocked port is preventing the next test to be finshed because `chromedriver-linux` then errors with the line seen in my first post above.

I did a screencast of the processes running a single test which works. Then again which doesn't work due to the blocking process. Then killing this process manually and running the test again successfully.

![dusk 1](https://user-images.githubusercontent.com/16897125/29718736-08413f78-89b4-11e7-9fa5-c80a6b5d8950.gif)

raniesantos:
I have the exact same problem. I need to keep killing the process to make dusk work again.

glatzor:
ProcessBuilder doesn't use the exec command to run the chromedriver. You should switch to pure Process. See https://github.com/symfony/symfony/issues/5759 for some background information.

I worte a small Trait to fix the issue in my project - you have to adapt the namespace and relative path to chromedriver:

[...]

PR #348, opened 2017-09-03
Use Process instead of ProcessBuilder to stop chromedriver afterwards

glatzor:
For backwards compatibilty reasons simfony's ProcessBuilder doesn't prepend
the exec command to the commandline. So only the PID of the wrapping
shell gets stored, but not the one of the chromedriver.

A solution is to use Process directly.

See https://github.com/symfony/symfony/issues/5759 for background
information.

Closes laravel/dusk#341

### DIFF
--- a/src/SupportsChrome.php
+++ b/src/SupportsChrome.php
@@ -4,7 +4,6 @@ namespace Laravel\Dusk;
 
 use RuntimeException;
 use Symfony\Component\Process\Process;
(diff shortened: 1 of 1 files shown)


deleugpn:
Did you test this on Windows machines? If not, I would like a chance to do so if that's okay with Taylor.

taylorotwell:
Re-send this when someone tests on Windows.

PR #351, opened 2017-09-05
Chrome Process refactored

deleugpn:
- Move the logic of building the Chrome process to it's own class;
- Add unit tests for building Chrome Process;
- Implement @glatzor solution for #341 only for Linux and Mac;

### DIFF
--- a/src/Chrome/ChromeProcess.php
+++ b/src/Chrome/ChromeProcess.php
@@ -24,6 +24,11 @@ class ChromeProcess
         }
     }
 
+    /**
+     * Build the process to run Chromedriver.
+     *
+     * @return \Symfony\Component\Process\Process
+     */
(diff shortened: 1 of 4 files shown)


taylorotwell:
What are we trying to solve here?

deleugpn:
Symfony's ProcessBuilder doesn't guarantee that the PID of the process will be correctly stored, so you can't kill the process properly when calling `$process->stop();`. When using `Process` component directly, Symfony adds `exec` to the command, which makes it less platform-agnostic (doesn't work on Windows), but guarantee sigkill will be able to kill the process. 

This is an attempt to not break anything while fixing the ability to properly kill chromedriver properly using Symfony `Process`.

An alternative is to drop Windows support from Dusk.

PR #363, opened 2017-09-11
Add global wait time default on Browser.php

calebporzio:
I'm working on some browser tests that rely on micro-services that sometimes take longer than 5 seconds to respond.

Currently, all the "wait" methods hard-code a default of 5 seconds wait time.

This PR adds the ability to globally specify the wait time using:

`$browser::$waitInSeconds = X`

Now, `$browser->waitFor('selector', ...` will wait for the default if now time is explicitly passed in.

Thanks!
Caleb

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -62,6 +62,13 @@ class Browser
     public static $userResolver;
 
     /**
+     * The default wait time in seconds.
+     *
+     * @var int
+     */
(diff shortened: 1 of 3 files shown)


taylorotwell:
Can you check the failing tests?

calebporzio:
@taylorotwell - the failing tests were fixed by #362 - I updated this branch to include the fix.

PR #364, opened 2017-09-11
extend base .env values instead of redefining in .env.dusk.local

austenc:
This allows you do define only what you need to override in `.env.dusk.local` instead of having to copy over the whole `.env` file to the dusk one. Achieved by appending the _original_ `.env` file contents to the temporary one after `.env.dusk.local` is copied to `.env`.  **Dotenv will load the first key in case of duplicates.** This makes `.env.dusk.local` behave more like the `phpunit.xml` env overrides for unit tests.
 
Happy to discuss alternate methods or other ideas for implementing this. Thoughts?

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -164,6 +164,9 @@ class DuskCommand extends Command
         copy(base_path('.env'), base_path('.env.backup'));
 
         copy(base_path($this->duskFile()), base_path('.env'));
+
+        $base = file_get_contents(base_path('.env.backup'));
+        file_put_contents(base_path('.env'), "\n" . $base, FILE_APPEND);
     }
 
     /**
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -166,7 +166,8 @@ class DuskCommand extends Command
         copy(base_path($this->duskFile()), base_path('.env'));
 
         $base = file_get_contents(base_path('.env.backup'));
-        file_put_contents(base_path('.env'), "\n" . $base, FILE_APPEND);
+
+        file_put_contents(base_path('.env'), PHP_EOL.$base, FILE_APPEND);
     }
 
     /**


austenc:
Awesome, thanks @taylorotwell!

KristofMorva:
@austenc @taylorotwell it's broken this way.
> By default, Dotenv will NOT overwrite existing environment variables that are already set in the environment.

It's true, however, that's not how Laravel Dusk works.
> If you want Dotenv to overwrite existing environment variables, use overload instead of load

And in `DuskCommand.php`: `(new Dotenv(base_path()))->overload();`

So currently always the **latest** variable will be loaded, which means, `.env` is loaded instead of `.env.dusk`.

austenc:
@KristofMorva the `.env` file in the line you referenced is actually a combination of both files, it ends up like this:

```
// This is from .env.dusk.*
SOME_VAR='123'

// This is from the main .env
SOME_VAR='456'
```

Dotenv sees the top one and assigns `SOME_VAR` to `123`. If you look at the `backupEnvironment()` method you'll see that the following happens:
 - `.env` is copied `.env.backup`
- `.env.dusk.*` is copied to `.env`
- The contents of the original file (now in `.env.backup`) are appended to the new `.env` file

Does that cause a problem for you? Could you be a little more specific about what's happening in your case?

KristofMorva:
@austenc I understand the code, the problem is, dotenv **does** override `SOME_VAR` to `456`. Please check the attached code and quote.

austenc:
@KristofMorva since the resulting file (after copy/append) looks like the example from my most recent comment, that call to `overload()` only picks up the first key that exists in the file. When parsing a single file, this is how dotenv behaves in the event of a duplicate variable/key. You might try testing this functionality in your own app by putting something like this in your main `.env` file -- you don't even need dusk to see how it works:
```
SOME_VAR='123'
SOME_VAR='456'
```
Then try a `dd(getenv('SOME_VAR'))` in a route or something and you should get '123', correct?

Testing this locally, I cannot get it to load `456` if that is defined in my main `.env` and overridden in `.env.dusk.*`. Do you have a sample project we could take a look at? 
Also, what version of `vlucas/phpdotenv` do you have installed?

austenc:
@KristofMorva are you using Valet, Homestead, or something else? More specifically, which web server are you using?

deleugpn:
I haven't tested this yet, will do as soon as I can, but I would like to point out that this is extremely dangerous for companies that share a single dev environment database across developers.

Dusk command has one environment and the browser may or may not have another (depending on whether you get the settings correctly). That said, it's common practice to build up and tear down databases for test purposes. If for some reason I forget to override this variable in my dusk env file, I could blow up my dev database easily. Sure, it's not as bad as blowing up production, but it would still be a pain.

Phpunit is a lot less likely to fall on this because we can always refer to in memory sqlite.

austenc:
Hey @deleugpn -- maybe this is better as some sort of configurable setting like  `appendsEnv` then? What may work for one use-case doesn't always work for another, for sure. Can definitely see some merit in supporting both types of env loading.

KristofMorva:
@austenc in generic Laravel, it should be fine. But with Dusk, `overload()` gets the **last** variable, not the first - that's how **load** works. I don't have a custom phpunit package fetched, I'm using whichever is in Laravel 5.5 (I'm using Docker btw).

I had `APP_ENV=local` in my `.env`, and `APP_ENV=testing` in my `.env.dusk`.  Then, this PR composed them to:
```
APP_ENV=testing
APP_ENV=local
```

Until here, it's what you'd expect. However, based on the `README` of `phpdotenv`, `overload()` will enable overwriting, therefore `APP_ENV=local` will be applied. I have also tested it, `App::getEnvironment()` prints `local`.

austenc:
@KristofMorva Yep, I'm testing in dusk. This is a shot in the dark, but did you try a different key other than `APP_ENV`?

KristofMorva:
@austenc I haven't, I have only experienced, after updating to this, my tests were failing, I have checked, the environment was wrong. Then I checked in the phpdotenv documentation, and I have seen with `override` the last one is loaded. What I'm thinking, maybe for the first test `.env` is loaded with `load`. Then, when preparing for the next test, `override` is used (that's how it seems in `DuskCommand.php`). If you check it out with multiple Dusk files and multiple tests, is it broken for you too?

austenc:
@KristofMorva I finally see what's happening. This causes the browser instance to potentially have a different environment than the dusk test file itself.

austenc:
@taylorotwell we'll need to figure out another solution to this. It's like we need the config to be in one order for dusk, and another by the time the browser instance is spawned to get consistent behavior.

PR #365, opened 2017-09-12
Revert "extend base .env values instead of redefining in .env.dusk.local"

deleugpn:
Reverts laravel/dusk#364


austenc:
@taylorotwell are you opposed to making this configurable? I think we can accomplish the 'merging' idea in a consistent way, but as @deleugpn noted in #364, there may be some who wish to toggle that functionality on and off.

See here: https://github.com/laravel/dusk/compare/2.0...austenc:2.0

Seems like people might have use for both approaches?

PR #367, opened 2017-09-13
Add scrollTo method

BryanHeath:
This will scroll the screen using JQuery's animate to an element matching the given selector.  

This can help with unknown error: Element is not clickable at point.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -224,6 +224,19 @@ class Browser
 
         return $this;
     }
+    
+    /**
+     * Scroll screen to element at the given selector.
+     *
+     * @param  string  $selector
+     * @return $this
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'll let this be macro'd in for now.

dotancohen:
Considering that Dusk cannot access elements not in the viewport, what is the reason for not merging in a commit that enables the code to scroll to the proper place? I'm not arguing, I'm genuinely interested as I see this as a critical feature.

mbardelmeijer:
To everyone wondering: this has been re-submitted under https://github.com/laravel/dusk/pull/684 and merged in the 5.0 branch. 🎉

PR #369, opened 2017-09-13
Have a manual override to enable Dusk in production.

webpatser:
An option to disable the production-mode check would be a nice addition. Sometimes the security risks are not valid if you know what you are doing.

Dusk is an awesome tool, not only for testing websites 👍

### DIFF
--- a/src/DuskServiceProvider.php
+++ b/src/DuskServiceProvider.php
@@ -39,7 +39,7 @@ class DuskServiceProvider extends ServiceProvider
      */
     public function register()
     {
-        if ($this->app->environment('production')) {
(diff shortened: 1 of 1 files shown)


webpatser:
All tests should be green. Double checked locally, for some reason a 7.0 test timed-out on travis. the 7.1 did return all green.

deleugpn:
I don't think you can use `env` helper here.

webpatser:
The env() seems to work fine.

taylorotwell:
Can't use env() helper outside of configuration files as noted earlier.

webpatser:
if I move the env helper to the dusk config file. And use that. Is it OK then?

PR #376, opened 2017-10-02
Overridable Chrome Driver Arguments via Env

taufek:
**Why we need this?**

In some setup, it needs to have different chrome driver arguments in order to run Dusk tests.
For instance, in laradock workspace container, the default argument (`--headless`), does not work.
I need to be able to change that to `--sandbox` to make it work.

**What has changed?**

Get the chrome driver arguments from env and maintain the previous arguments as default values.


taylorotwell:
You own this class, so you can change it however you want.

PR #378, opened 2017-10-02
Test Cases for MakeAssertions

taufek:
Add more test cases for MakeAssertions concern.


taylorotwell:
I don't usually test getters.

PR #384, opened 2017-10-08
Add option to disable direct tty output

antonioribeiro:
Setting 

``` php
$process->setTty(true);
```

Prevents output from being piped. So if we do 

``` 
php artisan dusk > piped.txt

# or

php artisan dusk | tee piped.txt

# or even

php artisan dusk > file1.txt 2>&1 | tee file2.txt
```

We get an empty files. This PR adds an option to disable output directly to tty.

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -15,7 +15,7 @@ class DuskCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'dusk';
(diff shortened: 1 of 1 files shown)


taylorotwell:
Maybe we should just not enable TTY at all then rather than have this option to disable it? What is the advantage of enabling it at all?

antonioribeiro:
Terminal colors, we would loose all those reds and greens...

deleugpn:
If we ever want to implement another argument on Dusk, it's not going to be trivial, will it?

PR #391, opened 2017-10-19
Updating chromedrivers to version 2.33

tnylea:
I still have some dusk tests running in 1.0 and Laravel 5.4. I was running into some issues where the chromedriver would not scroll to a certain position, after I updated the chromedrivers to the latest version in the bin folder, this solved my problem.

Wondering if we can tag a new version of 1.1.1, so that way anyone else with this issue can resolve it by doing a composer update.

Thanks guys 👍

PR #398, opened 2017-10-24
Updating chrome drivers

tnylea:
Hey @taylorotwell 

I've made sure to change the permissions back on the Windows Chromedriver, this is in reference to this PR: https://github.com/laravel/dusk/pull/391

Not sure why the permissions changed on that file, but this PR has updated that. Let me know if you need anything else and if we can get these merged into the version 1.0 branch.

Appreciate it 👍


deleugpn:
This has already been merged on https://github.com/laravel/dusk/pull/395 for version 2.0

But we might have problems with these types of updates. The scenario: 

- Use Dusk 2.0.x with Chrome Version 2.33 and Google Chrome 62
- 9 months from now Dusk 2.0.x updates to Chrome Version 2.40 and Google Chrome 68
- A simple `composer update` will grab a new Chrome Version, but if you're still on Chrome 62, it will break. 

How to avoid the breaking change without releasing a new major patch every chrome version update?

chilio:
@deleugpn 
One solution would be to put chromedriver builds to different repo for pulling exact version via composer.
I think that this approach might reduce the maintenance a bit and separate chromedriver releases from dusk ones...    
Then in dusk start procedure...
1. try to start system inbuilt chromedriver (probably it exists and is in correct version already),
2. if not then (as a last resort) present info to the user with options:
- installing chrome driver manually
or
- requiring specific version of chromedriver for current system via composer and modyfing DuskTestCase config (this might be automated, but for the start simple info would be enough)

tnylea:
@deleugpn This is for version 1.0.

Apps that are still running on 5.4 cannot use 2.0, can you please re-open :)

Thanks.

PR #400, opened 2017-10-25
Add assertBodyHasClass assertion

przemdz665:
Simple assertion to check if body element has given class name.


taylorotwell:
No plans on adding. You can macro it in.

PR #425, opened 2017-12-29
Replace ProcessBuilder with Process and remove redundant method.

crynobone:


deleugpn:
Did you teste this on Windows? The reason I did this way was exactly to keep Windows Support.

crynobone:
ProcessBuilder has been removed in Symfony 4.0, since the only option is to use Process class directly it make sense to just use the `process()` method.

PR #429, opened 2018-01-07
[2.0] Missing type hinting

carusogabriel:


deleugpn:
Perhaps you meant `mixed`?

carusogabriel:
Taylor uses `dynamic` in the core, when we **really** do not know what is going to be ☺️

PR #430, opened 2018-01-07
 Allow tests to be run on an alternative port

duncan3dc:
By default the `chromeprocess` binaries run on port 9515, but they accept a command line option to use an alternative port, this PR makes use of that to allow tests to use whatever port they like.

The use case for this is to run tests in parallel, allowing us to run one test class on port 9515 and another on 9516 simultaneously:
```php
{
    $port = 9516;
    static::startChromeDriver($port);
    static::setPort($port);
}
```

### DIFF
--- a/src/Chrome/ChromeProcess.php
+++ b/src/Chrome/ChromeProcess.php
@@ -17,14 +17,23 @@ class ChromeProcess
     protected $driver;
 
     /**
+     * The port to run the Chromedriver on.
+     *
+     * @var int
+     */
+    protected $port;
+
+    /**
      * Create a new ChromeProcess instance.
      *
      * @param  string  $driver
+     * @param  int     $port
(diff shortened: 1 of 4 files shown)


duncan3dc:
Thanks @taylorotwell :heart:

taylorotwell:
Literally breaks Dusk entirely. Sigh.

taylorotwell:
Reverted entirely. Dusk would not even run. PLEASE start testing your code at least manually.

duncan3dc:
Sorry about that, the code works perfectly for my purposes and is fully tested.

As all the unit tests passed here I assumed it was also compatible with your intended usage.

I'll take a look at how you use Dusk and see if I can add some unit tests to prevent breaking changes like this slipping through your CI in future

PR #445, opened 2018-01-31
Fix return value

helmut:
Fixed typo for`assertHasQueryStringParameter`method returning $output rather than $this.


helmut:
No exactly sure why travis build failed. Unrelated to my changes I assume. Can an admin trigger it again?

PR #446, opened 2018-01-31
Basic mouse interactions.

monas:
These actions can be used to controll the mouse without selecting the element first. After making some regular action like $browser->mouseover('.some_selector') the mouse gets position of the center coordinates of that selected element. And for example to click away form some opened notification or dialog You could then use mouseMoveByOffset(-100, 100)->mouseClick(). Another use case could be interaction with some sliders (this was our issue). And even more - it would be possible to make pixel perfect assertions.

### DIFF
--- a/src/Concerns/InteractsWithMouse.php
+++ b/src/Concerns/InteractsWithMouse.php
@@ -35,4 +35,57 @@ trait InteractsWithMouse
 
         return $this;
     }
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Missing @returns on some doc block and uses snake_case whereas all of Laravel uses camelCase variables at this time.

PR #449, opened 2018-02-03
Allow tests to be run on an alternative port

duncan3dc:
Following on from #430 and #441 this pr introduces the alternative port feature:

```php
    /**
     * Prepare for Dusk test execution.
     *
     * @beforeClass
     * @return void
     */
    {
        $port = 9517;
        static::startChromeDriver($port);
        static::usePort($port);
    }
```

### DIFF
--- a/src/Chrome/ChromeProcess.php
+++ b/src/Chrome/ChromeProcess.php
@@ -17,13 +17,22 @@ class ChromeProcess
     protected $driver;
 
     /**
+     * The port to run the Chromedriver on.
+     *
+     * @var int
+     */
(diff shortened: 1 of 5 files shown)


duncan3dc:
I've tried it out with a fresh "composer create-project laravel/laravel" install, using the default port, and then using a custom port, the example test passes in both instances

taylorotwell:
No plans to add right now.

duncan3dc:
Would you be able to explain why?
Would you consider just accepting 19e75f2959a32843fc57955f68e7118338dbe298?

Issue #455, opened 2018-02-09
"Cannot use object of type Laravel\Dusk\Browser as array" error


iadambrown:
It seems that the following commit: https://github.com/laravel/dusk/commit/fb67e61819fb9bd23c363c2b49aa861ec1d9c724 now breaks the `assertQueryStringHas` assertion, as in the following code, `assertHasQueryStringParameter` now returns `$this`, which is an instance of the Laravel\Dusk\Browser object, instead of the `$output` array that is being tested against:

```
public function assertQueryStringHas($name, $value = null)
{
    $output = $this->assertHasQueryStringParameter($name);

    if (is_null($value)) {
        return $this;
    }

    PHPUnit::assertEquals(
        $value, $output[$name],
        "Query string parameter [{$name}] had value [{$output[$name]}], but expected [{$value}]."
    );

    return $this;
}
```

So the following line would now cause the "Cannot use object of type Laravel\Dusk\Browser as array" error:
`$browser->assertQueryStringHas('response_type', 'code');`

PR #456, opened 2018-02-09
Revert fb67e61 and fix return type

staudenmeir:
Fixes https://github.com/laravel/dusk/issues/455

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -223,7 +223,7 @@ trait MakesAssertions
      * Assert that the given query string parameter is present.
      *
      * @param  string  $name
-     * @return $this
+     * @return array
      */
     protected function assertHasQueryStringParameter($name)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't understand why all other assertions would return $this except for this one?

staudenmeir:
Please reopen @taylorotwell.
fb67e61 broke the `assertQueryStringHas()` assertion when testing a `$value`.
The changed method is just a helper method for the "real" assertion which returns `$this` as expected.

iadambrown:
Yes, `assertQueryStringHas()` calls `assertHasQueryStringParameter()`, but expects the `$output` array to assert against (`PHPUnit::assertEquals($value, $output[$name] ...);`).

However, since the `assertHasQueryStringParameter()` assertion is *also* listed in the docs, this could cause issues for those who are expecting `$this` to now be returned... so returning the array satisfies `assertQueryStringHas()`, but returning `$this` allows for further method chaining when calling `assertHasQueryStringParameter()` directly.

staudenmeir:
Good point about the docs. But as a protected method you can't call it from the outside...

Shouldn't we just merge `assertHasQueryStringParameter()` into `assertQueryStringHas()`? It doesn't provide any additional functionality as a separate method.

PR #461, opened 2018-02-13
fixes assertQueryStringHas

devinfd:
Due to https://github.com/laravel/dusk/pull/457, assertQueryStringHas was throwing a ```Cannot use object of type Laravel\Dusk\Browser as array``` Error

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -180,12 +180,14 @@ trait MakesAssertions
      */
     public function assertQueryStringHas($name, $value = null)
     {
-        $output = $this->assertHasQueryStringParameter($name);
+        $this->assertHasQueryStringParameter($name);
 
         if (is_null($value)) {
             return $this;
         }
-
+        
+        parse_str(parse_url($this->driver->getCurrentURL())['query'], $output);
+        
         PHPUnit::assertEquals(
             $value, $output[$name],
             "Query string parameter [{$name}] had value [{$output[$name]}], but expected [{$value}]."


staudenmeir:
#457 (should have) fixed that error. What call of `assertQueryStringHas()` causes it?

devinfd:
assertQueryStringHas is calling assertHasQueryStringParameter which you correctly changed to return $this, however it was returning a ```parse_str``` $output array that assertQueryStringHas was using to make further assertions against. This change has assertQueryStringHas responsible for parsing the current url query string independent of assertHasQueryStringParameter.

staudenmeir:
fb67e61 falsely changed the return value to `$this`, I changed it back to `$output` in #457.

devinfd:
The return value from assertHasQueryStringParameter has changed so many times in the past week I've lost track of what it currently is. I also just realized there is now a v3 branch of Dusk. So the ```Cannot use object of type Laravel\Dusk\Browser``` error actually only exists in the v2 branch.

But lets think about this. Why should assertHasQueryStringParameter return an array and not $this? That makes it no longer fluent and behave different from all the other assertions. Perhaps this PR should stand and assertHasQueryStringParameter should be reverted.

staudenmeir:
You're right, I forgot to fix it in the v2 branch. I'll create a PR.

I agree that using `$output` from a different method is not ideal. But since `assertHasQueryStringParameter()` is just a protected helper for `assertQueryStringHas()`, it doesn't have to be fluent.

Another possible solution could be to just merge `assertHasQueryStringParameter()` into `assertQueryStringHas()`. What do you think?

devinfd:
I guess I don't understand why assertHasQueryStringParameter is protected. Either it should be public or this PR should be closed and you PR to fix the v2 branch. Perhaps @taylorotwell should decide since there are many moving parts to this.

staudenmeir:
It could have been a mistake. But I dont't think it has to be public since it doesn't provide any additional functionality. Calling assertQueryStringHas() gives you the same result.

I already removed it from the docs (https://github.com/laravel/docs/pull/4060).

taylorotwell:
What in the heck is going on with this method?! Can we just revert things back to how they were before this madness started?

taylorotwell:
I think I'm just going to have to solve this myself because we're not getting anywhere at all.

taylorotwell:
I'm not merging anything related to this without tests.

PR #469, opened 2018-02-26
Update chrome binaries

themsaid:


gemal:
can this be backported to version 2 branch?

InstanceOfMichael:
I might be jumping the gun on this, but it seems like a quick temporary work-around for version 2 is to download the new version of the driver you need from https://github.com/themsaid/dusk/blob/705b3901ec204919c1bc7bb730c891cf0bc5b6b9/bin/ and overwrite the existing one in your `vendor/laravel/dusk/bin` folder. If you need it for CI/CD, it's an easy one-liner to add to your build.

PR #471, opened 2018-02-27
Prevent screenshot and console filename collisions

### LINKED ISSUE #352: Screenshot and Console Filename Collision
Screenshot and Console files are overwritten if multiple test classes have functions with the same names.

For example:
`class FooTest extends DuskTestCase {
    public function testExample() {}
}`
would produce a screenshot file named `failure-testExample-0.png`

`class BarTest extends DuskTestCase {
    public function testExample() {}
}`
would also produce a screenshot file named `failure-testExample-0.png` causing the first one to be overwritten.

Would it be better to use the `toString()` method instead of the `getName()` method to generate the filename? This would require cleaning the `toString()` result to remove any special characters but it would then provide a unique filename to stop the filename collision.

staudenmeir:
Adds class to filename to prevent overwritting files when multiple tests have the same name.

Fixes #352.

### DIFF
--- a/src/Concerns/ProvidesBrowser.php
+++ b/src/Concerns/ProvidesBrowser.php
@@ -132,7 +132,9 @@ trait ProvidesBrowser
     protected function captureFailuresFor($browsers)
     {
         $browsers->each(function ($browser, $key) {
-            $browser->screenshot('failure-'.$this->getName().'-'.$key);
+            $name = str_replace(['\\', ':'], '_', $this->toString());
+
+            $browser->screenshot('failure-'.$name.'-'.$key);
         });
     }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you give more information on what this PR does?

staudenmeir:
Suppose you have `Tests\FooTest::testExample()` and `Tests\BarTest::testExample()`.

If both tests fail, they each create a screenshot named `failure-testExample-0.png` – so the second screenshot overwrites the first one.

This PR adds the class to the filename to prevent these collisions:
`failure-Tests_FooTest__testExample-0.png`
`failure-Tests_BarTest__testExample-0.png`

The same goes for the console logs (without the `failure-` prefix).

taylorotwell:
I can't find a single method in Dusk named `toString` where is `$this->toString()` defined?

staudenmeir:
It's a PHPUnit method @taylorotwell

PR #472, opened 2018-02-27
Prevent screenshot and console filename collisions

staudenmeir:
(re-post of closed #471)

Answer to your question:
`toString()` is a method in `PHPUnit\Framework\TestCase` (as is the currently used `getName()`).


taylorotwell:
Reverted because it breaks things.

PR #475, opened 2018-03-07
Replace '/' from captured logs/screenshots names to prevent failure

pdbreen:
When running a test with a data provider that was supplying URLs, I hit an error because of an invalid file name being passed to file_put_contents.  This PR extends the filename scrubbing to remove the '/' which was causing the failure (at least on Windows).

### DIFF
--- a/src/Concerns/ProvidesBrowser.php
+++ b/src/Concerns/ProvidesBrowser.php
@@ -132,7 +132,7 @@ trait ProvidesBrowser
     protected function captureFailuresFor($browsers)
     {
         $browsers->each(function ($browser, $key) {
-            $name = str_replace(['\\', ':', '/'], '_', $this->toString());
+            $name = str_replace(['\\', ':', '/', '>', '<', '"', '|', '?', '*'], '_', $this->toString());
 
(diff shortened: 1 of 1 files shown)


staudenmeir:
Maybe we should just remove all data from the filename. It could contain all kinds of problematic characters and also lead to very long filenames.

pdbreen:
If the data provider index is retained, the file names would remain unique for each iteration.  Otherwise, only last error in set would survive.   Having the data value could be helpful, but not probably not a deal breaker for anyone if dropped.

If I had done a little homework before submitting the PR, this doc - https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx - gives the complete list of reserved characters for Windows.  I believe this list is a superset of linux restrictions, but I'm not an expert.

> The following reserved characters:
> 
> < (less than)
> > (greater than)
> : (colon)
> " (double quote)
> / (forward slash)
> \ (backslash)
> | (vertical bar or pipe)
> ? (question mark)
> * (asterisk)

taylorotwell:
This is getting kinda crazy.

taylorotwell:
Reverted the PR that caused this.

PR #477, opened 2018-03-07
Prevent breaking console and screenshot filenames and restore #472

staudenmeir:
The reverted PR #472 only caused part of the problem discussed in #475, the general problem with datasets existed before. Breaking filenames like  `testExample with data set "foo/bar"-0.log` can still occur.

This PR completely removes datasets from console and screenshot filenames.

It also restores the intended feature of #472 – adding the class to filenames to prevent collisions with duplicate test names:
`failure-Tests_FooTest_testExample-0.png`
`failure-Tests_BarTest_testExample-0.png`

### DIFF
--- a/src/Concerns/ProvidesBrowser.php
+++ b/src/Concerns/ProvidesBrowser.php
@@ -132,7 +132,7 @@ trait ProvidesBrowser
     protected function captureFailuresFor($browsers)
     {
         $browsers->each(function ($browser, $key) {
-            $browser->screenshot('failure-'.$this->getName().'-'.$key);
+            $browser->screenshot('failure-'.$this->getName(false).'-'.$key);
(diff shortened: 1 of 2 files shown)


taylorotwell:
How do I know this isn't broken too?

staudenmeir:
We now only add the test's class and method name to the filename.
Both are limited by this regular expression: `[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*`
I successfully tested filenames with all possible characters on Linux and Windows.
Namespace separators are replaced with underscores.

Should I add tests?

taylorotwell:
You should probably add tests, yes.

staudenmeir:
I added tests.

PR #489, opened 2018-03-15
Update chrome binaries

JesseLeite:
Updating broken chrome binaries, again.  Should be tested on all OSes, and backported.

Is there a way to automate this somehow, so that it doesn't require a PR to dusk every time?


browner12:
FYI, PR's with binaries will not be accepted. The core team are the only people allowed to update the binaries, for security reasons.

JesseLeite:
Very fair!  Closing this.  My question still stands though:

>Is there a way to automate this somehow, so that it doesn't require a PR to dusk every time?

Like, if the binaries were extracted out to a separate package, and someone were really on top of keeping that up-to-date, then packages like dusk wouldn't have to worry about it.  Thoughts? @themsaid?

browner12:
I believe there was talk of making an Artisan command that users could run to download the binaries, so we weren't reliant on the package for this.

Where do you download these binaries? I've found a couple places, and I'm curious where the official source is.

robindrost:
offical repo:

https://chromedriver.storage.googleapis.com/index.html?path=2.36/

staudenmeir:
I proposed an Artisan command: https://github.com/laravel/dusk/pull/492

amitbk:
SOLVED!!!

In **DuskTestCase.php** File

**Old Code of driver():**

```
protected function driver()
    {
        $options = (new ChromeOptions)->addArguments([
            '--disable-gpu',
            '--headless',
            '--window-size=1920,1080',
        ]);

        return RemoteWebDriver::create(
            'http://localhost:9515', DesiredCapabilities::chrome()->setCapability(
                ChromeOptions::CAPABILITY, $options
            )
        );
    }
```
Just Add **'--no-sandbox'** flag

**New Code:**

```

protected function driver()
    {
        $options = (new ChromeOptions)->addArguments([
            '--disable-gpu',
            '--headless',
            '--window-size=1920,1080',
            '--no-sandbox'
        ]);

        return RemoteWebDriver::create(
            'http://localhost:9515', DesiredCapabilities::chrome()->setCapability(
                ChromeOptions::CAPABILITY, $options
            )
        );
    }
```

Issue #490, opened 2018-03-16
Cannot update dusk w/ new chromedriver in Laravel 5.4 app


JesseLeite:
I realize we're behind on Laravel 5.4, but we cannot update laravel/dusk in our app any more without illuminate/console constraint error.  I -think- we just need updated chromedriver, nothing new in the code itself?

I've heard talks of an artisan command to update dusk's chromedriver, regardless of laravel version or dusk version.  Thoughts?

PS. Not complaining.  Maybe this will push us to finally upgrade to 5.6 🙈

staudenmeir:
What's the error message?

JesseLeite:
I've done that already, but should users have to manually mess with vendor files?

staudenmeir:
Only if they use unsupported Laravel versions ;-) Does it work?

JesseLeite:
Yezzir.

staudenmeir:
I created an Artisan command for updating the ChromeDriver binaries on all Dusk versions: https://github.com/staudenmeir/dusk-updater

driesvints:
Heya, really sorry about this but unfortunately we cannot provide support for Laravel 5.4 anymore. Please try upgrading or the solution proposed by @staudenmeir.

ryancwalsh:
@staudenmeir What a great idea! Thanks.

PR #492, opened 2018-03-18
[Proposal] Artisan command for updating ChromeDriver

staudenmeir:
This PR proposes an Artisan command for updating the ChromeDriver binaries:

    php artisan dusk:update

It can be used to immediately get the latest ChromeDriver version after its release.
People with Laravel 5.4 can use it to keep their Dusk 1.* installation alive (#490).
By running it after `composer install/update` it could also replace shipping the binaries with Dusk altogether.

I tested it on Linux and Windows. Can someone please test it on a Mac?

The command requires the [PHP Zip extension](http://www.php.net/manual/en/book.zip.php).

It currently downloads the binaries for all three operating systems.
I thought about only updating the ChromeDriver for the running OS. But there could be cases where people run Homestead on Mac/Windows and accidentally execute the update command on the Host OS. Then they would still be using the old ChromeDriver in Homestead.
We could also add the option to specify the OS (e.g. `--linux`).

Another improvement could be to compare the installed ChromeDriver version with the latest release to avoid unnecessary downloads. I just wanted to get some feedback before implementing those.

### DIFF
--- a/src/Console/UpdateCommand.php
+++ b/src/Console/UpdateCommand.php
@@ -1,0 +1,139 @@
+<?php
+
+namespace Laravel\Dusk\Console;
+
+use Illuminate\Console\Command;
+use ZipArchive;
+
+class UpdateCommand extends Command
+{
+    /**
+     * URL to the latest release version.
+     *
+     * @var string
+     */
+    public static $versionUrl = 'https://chromedriver.storage.googleapis.com/LATEST_RELEASE';
+
+    /**
+     * URL to the ChromeDriver download.
+     *
+     * @var string
+     */
+    public static $downloadUrl = 'https://chromedriver.storage.googleapis.com/%s/chromedriver_%s.zip';
+
+    /**
+     * Download slugs for the available operating systems.
+     *
+     * @var array
+     */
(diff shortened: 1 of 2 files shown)


deleugpn:
Should `composer.json` to either require the zip extension or at least suggest it?

staudenmeir:
If the binaries get removed from the repository and the command is the only way to get them, we could add an alternative like [PclZip](https://packagist.org/packages/pclzip/pclzip).

browner12:
I think the PHP Zip Extension is sufficient.  That's the same thing we use for the Laravel Installer.

https://github.com/laravel/installer/blob/master/src/NewCommand.php#L43

If we do add it to `composer.json`, I would say suggest it, don't require it.  We don't want to break functionality for users who are perfectly fine without this.

staudenmeir:
Good point. I guess the number of people without the extension is pretty low. In a dev environment you should be able to install it.

taylorotwell:
I'll just keep updating them for now.

staudenmeir:
If someone wants to use the update command, I created a Composer package: https://github.com/staudenmeir/dusk-updater

PR #497, opened 2018-03-31
add `waitForRoute` method

browner12:
this is a convenience method to use routes instead of URLs.

behaves the same as `visitRoute` as a wrapper of `visit`.

not sure if this is possible to unit test. couldn't find a test for `waitForLocation`, but let me know if one needs to be added.


taylorotwell:
Was missing doc block description. Added manually.

browner12:
great, thanks!

PR #503, opened 2018-04-07
Switch the phpunit to a required dependency

duncan3dc:
The [MakesAssertions](https://github.com/laravel/dusk/blob/3.0/src/Concerns/MakesAssertions.php) requires phpunit, but it's only declared as a dev dependency.

This PR switches it so that it's always available

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,12 +15,12 @@
         "nesbot/carbon": "~1.20",
         "illuminate/console": "~5.6",
         "illuminate/support": "~5.6",
+        "phpunit/phpunit": "~7.0",
         "symfony/console": "~4.0",
         "symfony/process": "~4.0"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


browner12:
FYI @duncan3dc . 99% of tests do not run in a production environment, so dev dependencies are installed in these cases.

If your project requires running Dusk and PHPUnit on a production machine, you can make this switch for your specific use.

deleugpn:
I think it's completely fair to say that `Dusk` have a `requirement` on **PHPUnit** to function. Dusk is a package intended to help the developer to run browser tests. The "production" environment for Dusk is people's computer / CI environments. Dev Dependencies are things required for developers to work on the project itself. Although Dusk requires phpunit for people to contribute to this package, it also requires it to function.

browner12:
ahh, gotcha. i see your point now.

PR #506, opened 2018-04-13
Customize title assertion error messages

browner12:
Provides clearer error messages for failed assertions.


staudenmeir:
There's "not" missing in the message.

browner12:
thanks for the catch @staudenmeir !

PR #515, opened 2018-05-07
[3.0] Add assertion tests

staudenmeir:
Adds the remaining assertion tests.

Now without `$resolver` mocks and with as few mocks as I could achieve.


taylorotwell:
It's all just mocked. It's not necessary.

PR #521, opened 2018-06-01
Add support for passing arguments to the chromedriver process.

tomzx:
This is quite useful to add support for running multiple instances of Dusk in parallel (by setting the `port` argument for instance).

### DIFF
--- a/src/Chrome/ChromeProcess.php
+++ b/src/Chrome/ChromeProcess.php
@@ -33,36 +33,38 @@ class ChromeProcess
     /**
      * Build the process to run Chromedriver.
      *
+     * @param array $args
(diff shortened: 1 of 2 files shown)


browner12:
could you provide an example of how this works?  is this something you call from the command line?

tomzx:
@browner12 

In your `DuskTestCase.php` file, you can have the following for example:

```php
{
	static::startChromeDriver([
		'--port='.env('DUSK_PORT', 9515),
		'--silent',
	]);
}
```

browner12:
great thanks.  hoping this was going to let us pass arguments from the command line into the `DuskTest`, but we'll have to keep working on that one.

Issue #464, opened 2018-02-14
Error: Call to undefined method Facebook\Webdriver\Remote\RemoteWebDriver::getCapabilities()


zoispag:
Since the Chrome update to v64 this week, my test that are uploading a file as part of a modal, are failing.
Tests are running through TeamCity on ubuntu 16.04

I've tried several ChromeDriver version including 2.34 & 2.35

Anyone experiencing something similar?

zoispag:
This comes from the **attach** method in InteractsWithElements trait:

```php
    /**
     * Attach the given file to the field.
     *
     * @param  string  $field
     * @param  string  $path
     * @return $this
     */
    public function attach($field, $path)
    {
        $element = $this->resolver->resolveForAttachment($field);

        if ($this->driver->getCapabilities()->getBrowserName() == 'phantomjs') {
	        $element->setFileDetector(new UselessFileDetector())->sendKeys($path);
        } else {
	        $element->setFileDetector(new LocalFileDetector)->sendKeys($path);
        }

        return $this;
    }
```
to check if the browser is phantomjs.

oncet:
I'm having the same error while trying the example test provided by Dusk 2.0.12.

```
namespace Tests\Browser;

use Tests\DuskTestCase;
use Laravel\Dusk\Browser;
use Illuminate\Foundation\Testing\DatabaseMigrations;

class ExampleTest extends DuskTestCase
{
    /**
     * A basic browser test example.
     *
     * @return void
     */
    public function testBasicExample()
    {
        $this->browse(function (Browser $browser) {
            $browser->visit('/')
                    ->assertSee('Laravel');
        });
    }
}
```

After running `php artisan dusk` I get:

```
1) Tests\Browser\ExampleTest::testBasicExample
Error: Call to undefined method Facebook\WebDriver\Remote\RemoteWebDriver::getCapabilities()
```
I'm using Laravel 5.5.34.

oncet:
All right, fixed after updating the facebook/webdriver package.

```
composer update facebook/webdriver
  - Updating facebook/webdriver (1.1.1 => 1.5.0): Downloading (100%)
```
`php artisan dusk` no longer throws the error and all is green.

pacoorozco:
I've the same problem, in order to reproduce it:
1. Install laravel and copy `.env.example`as `.env`
1. Run `composer require --dev laravel/dusk`
1. Run `php artisan dusk:install`
1. Run `php artisan key:generate`

[...]

PR #528, opened 2018-06-20
Update the requested facebook/webdriver version to 1.3

### LINKED ISSUE #385: Error: Call to undefined method Facebook\Webdriver\Remote\RemoteWebDriver::getCapabilities()
I upgraded my laravel vendor and run the tests, then got this error.

I then fall back the Facebook Webdriver and it works OK.

crhg:
Dusk now calls `Facebook\WebDriver\Remote\RemoteWebDriver::getCapabilities()` method which is added to facebook/webdriver since version 1.3.
(https://github.com/facebook/php-webdriver/blob/community/CHANGELOG.md)

Therefore, I think that the request version described in composer.json should be updated.

Fixes #385,#464

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": ">=7.1.0",
-        "facebook/webdriver": "~1.0",
+        "facebook/webdriver": "~1.3",
         "nesbot/carbon": "~1.20",
         "illuminate/console": "~5.6",
         "illuminate/support": "~5.6",


deleugpn:
~1.0 will load 1.3 already.

crhg:
This fix is to prevent incorrect installation of version older than 1.3 for some reason.

As said in #385, facebook/webdriver 1.2 is installed if php zip extension is not installed. But 1.2 has no `getCapabilities` method and will not work properly.

If you apply this fix, you will get an error like the following when installing, and it will be easy to see that it is a problem that zip extension is missing.

```console
# php composer.phar install
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - facebook/webdriver 1.6.0 requires ext-zip * -> the requested PHP extension zip is missing from your system.
    - facebook/webdriver 1.5.0 requires ext-zip * -> the requested PHP extension zip is missing from your system.
    - facebook/webdriver 1.4.1 requires ext-zip * -> the requested PHP extension zip is missing from your system.
    - facebook/webdriver 1.4.0 requires ext-zip * -> the requested PHP extension zip is missing from your system.
    - laravel/dusk dev-update-facebook-webdriver-version requires facebook/webdriver ~1.3 -> satisfiable by facebook/webdriver[1.3.0, 1.4.0, 1.4.1, 1.5.0, 1.6.0].
    ...
```

Currently incompatible versions are installed siliently, and at runtime the following error occurs:

```console
There was 1 error:

1) Tests\Browser\ExampleTest::testBasicExample
Error: Call to undefined method 
Facebook\WebDriver\Remote\RemoteWebDriver::getCapabilities()

/var/www/vendor/laravel/dusk/src/Browser.php:257
```

If you check the error message with google for example, you may be able to reach the solution, but it requires some effort.

PR #532, opened 2018-07-12
Make clickLink() only select visible elements

danmichaelo:
I think it makes sense that `clickLink()` only selects visible elements.

(Btw. should pull requests be made against `master` or `3.0`? Would be nice if this was stated somewhere, e.g. in the readme, since it's a bit confusing when `3.0` is the main branch.)


taylorotwell:
A breaking change like this would go to master.

Issue #320, opened 2017-07-19
Screenshot full page


xakzona:
How to screenshot full page? This issue did't help me
https://github.com/laravel/dusk/issues/279

furqanfreed:
I am having same issue.

`$response = $browser->visit( url for page with scroll on Y-axis);
$response->resize(1100, 2500)->driver->takeScreenshot($path);`


it will only take screen shot of browser visible area on screen, even if you make width and height dynamic using following 


`$element = $response->element('body');
$element_width = $element->getSize()->getWidth();
$element_height = $element->getSize()->getHeight();`

xakzona:
I added this method to DuskTestCase
```protected function captureFailuresFor($browsers)
    {
        $browsers->each(function (Browser $browser, $key) {
            $width  = $browser->driver->executeScript("return Math.max(document.body.scrollWidth, document.body.offsetWidth, document.documentElement.clientWidth, document.documentElement.scrollWidth, document.documentElement.offsetWidth);");
            $height = $browser->driver->executeScript("return Math.max(document.body.scrollHeight, document.body.offsetHeight, document.documentElement.clientHeight, document.documentElement.scrollHeight, document.documentElement.offsetHeight);");
            $size = new WebDriverDimension($width, $height);
            $browser->driver->manage()->window()->setSize($size);
            $browser->screenshot('failure-'.$this->getName().'-'.$key);
        });
    }```

**Comment by oncet** [comment]
I'm having the same issue, when a test fails I only get a cropped screenshot. Overriding the `captureFailuresFor()` method solves the problem. But why this isn't working properly by default? Could it be something in my environment?

I'm using Laravel 5.5 on Ubuntu.

**Comment by jhm-ciberman** [comment]
To the answer of @xakzona remember to add this to the top of the file:

```
use Laravel\Dusk\Browser;
use Facebook\WebDriver\WebDriverBy;
use Facebook\WebDriver\WebDriverDimension;
```

ahtinurme:
The modern, updated version:

$browsers->each(function (Browser $browser, $key) {
            $body = $browser->driver->findElement(WebDriverBy::tagName('body'));
            if (!empty($body)) {
                $currentSize = $body->getSize();
                $size = new WebDriverDimension($currentSize->getWidth(), $currentSize->getHeight());
                $browser->driver->manage()->window()->setSize($size);
            }
            $name = $this->getCallerName();
            $browser->screenshot('failure-' . $name . '-' . $key);
        });

PR #540, opened 2018-07-23
ScreenShotFull page

xakzona:
Related issue https://github.com/laravel/dusk/issues/320


taylorotwell:
I don't really like the idea of adding additional configuration options. Would rather have good default behavior if possible?

Minhyme:
Default behavior on failure is still to only screenshot visible area. Important debug information like url and download files bar are not visible.

PR #542, opened 2018-08-03
[3.0] Simplify switching frames

browner12:
If a page includes an iFrame, it is currently rather clunky to interact with it.  For example, we use a 3rd party form service to submit and aggregate things like Contact forms. I would like to test that we can indeed submit to them, and need to interact with their iFrame to do so.

this new method provides a seamless way to switch frames, and continue chaining methods on the browser.  Simply pass any valid selector to the `frame()` method.

```php

            $browser->visit('https://example.com/contact')
                    ->frame('#iframeId')
                    ->type('name', 'Andrew Brown')
                    ->type('email', 'andrew@example.com')
                    ->press('submit')
                    ->assertSee('Thank you for contacting us!');
        });
```

If you need to continue interacting with the default frame when done, you can call the method without an argument.

```php

            $browser->visit('https://example.com/contact')
                    ->frame('#iframeId')
                    ->type('name', 'Andrew Brown')
                    ->type('email', 'andrew@example.com')
                    ->press('submit')
                    ->assertSee('Thank you for contacting us!')
                    ->frame()
                    ->click('#defaultFrameElement');
        });
```

Credit for this code goes to everyone in this thread:

https://laracasts.com/discuss/channels/laravel/dusk-click-element-in-iframe?page=1

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -410,6 +410,23 @@ class Browser
     }
 
     /**
+     * Switch to a specified frame in the browser.
+     *
+     * @param  string  $selector
+     * @return $this
+     */
+    public function frame($selector = null)
+    {
+        if ($selector) {
+            $this->driver->switchTo()->frame($this->resolver->findOrFail($selector));
+        } else {
+            $this->driver->switchTo()->defaultContent();
+        }
+
+        return $this;
+    }
+
+    /**
      * Dynamically call a method on the browser.
      *
      * @param  string  $method


taylorotwell:
Should it be contained within a Closure or something maybe so that it automatically switches back to default frame?

```php
->withinFrame('selector', function ($browser) {})
```

browner12:
interesting idea.

so the `withinFrame()` method would look something like this?

```php
public function withinFrame($selector = null, $closure)
{
    $this->driver->switchTo()->frame($this->resolver->findOrFail($selector));

    $closure($this);

    $this->driver->switchTo()->defaultContent();

    return $this;
}
```

could this get a little clunky if people had to pass in variables to the closure with `use`?  for example, maybe some people would pass in models generated by their factories with fake data.

taylorotwell:
Yeah that's basically what it would look like. It's just more consistent with the existing `within` method, etc.

browner12:
yah, I'm good with this if it keeps things consistent.

On the `within` method, what is the reason for generating a new `Browser` instance, rather than using the existing one?

taylorotwell:
I don't remember. Maybe.

PR #573, opened 2018-10-29
ProvidesBrowser createBrowsersFor wrong browser count 

xakzona:
If 2 (and more) browsers declared in setUp method, test will open minumum one more window, because  count 1 is builded in this code `$this->browsersNeededFor($callback) - 1`


taylorotwell:
Please re-submit with a better explanation and a snippet of code we can use to recreate the issue.

PR #576, opened 2018-11-01
[4.0] Latest chromedriver (2.43)

alexbowers-testing:
Reason it's needed:

- Stripe checkout is based on iframes and there has been a change to Chrome that means that Chrome 70 (and perhaps earlier) does not work changing frames, meaning that a stripe checkout process can no longer be tested.

This will require a new tagged release.

I'd like to press for an expedited release if possible.

### DIFF
--- a/src/TestCase.php
+++ b/src/TestCase.php
@@ -41,7 +41,7 @@ abstract class TestCase extends FoundationTestCase
     protected function driver()
     {
         return RemoteWebDriver::create(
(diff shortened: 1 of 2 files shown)


alexbowers:
I'm out of ideas why the tests aren't working.

I have tried it in both Vagrant (ubuntu 16.04) and Mac, and the tests pass on both.

driesvints:
Maybe it's best that you try this on your own Travis build until you figure it out. Feel free to resend this once you do!

PR #584, opened 2018-12-03
fix environment check in provider

prolic:


deleugpn:
What bug does this fix?

prolic:
My bad, I only checked the interface, and there is no argument, but the implementation uses `if (func_num_args() > 0) {` - sorry, I close it.

Issue #570, opened 2018-10-23
Stripe Elements requires upgrade of ChromeDriver


alphametric:
Simple one here... 

Working with iframes led to a test failure (example in question was Stripe Elements related). Upgrading the ChromeDriver to the latest version solved the issue. My guess is it relates to the recent Google Chrome update.

alexbowers:
Unfortunately, this doesn't seem to be as easy as just updating the files.

Whilst that does work, and locally for me in Vagrant and on the mac the tests all pass, for some reason Travis isn't having it.

https://github.com/laravel/dusk/pull/576

If anybody can figure that out, that'd be great.

staudenmeir:
@alexbowers Is Travis using the latest Chrome version?

staudenmeir:
You can try my updater package: https://github.com/staudenmeir/dusk-updater

Install it with `composer require --dev staudenmeir/dusk-updater` and then add `php artisan dusk:update` to your `.travis.yml` file.

I just tested it on Travis, I can get the latest Chromedriver binaries. I verified it with `echo exec(base_path('vendor/laravel/dusk/bin/chromedriver-linux --v'));`.

davidhemphill:
I just had this same issue working on a site, and updating the ChromeDriver binaries solved it. Thanks 👍

driesvints:
Sent in a PR which updates the chromedriver to 2.45: https://github.com/laravel/dusk/pull/591

driesvints:
@staudenmeir used your package for this. Worked really well. Thanks for this!

PR #591, opened 2018-12-18
[4.0] Update Chromedriver

driesvints:
Chromedriver was out of date which was causing some issues.

Fixes https://github.com/laravel/dusk/issues/570

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,6 @@ before_install:
 
 install: travis_retry composer install --no-interaction --prefer-source
 
+before_script: sudo sh -c 'echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6'
+
 script: vendor/bin/phpunit --verbose


driesvints:
Trying to figure out why the tests are failing.

driesvints:
Almost certain it's related to this Travis issue with IPv6 not being enabled: https://github.com/travis-ci/travis-ci/issues/8891

tentus:
This has turned out to be a breaking change for my team, but frustratingly, nailing down exactly what's actually going on has eluded us. 

Specifically on Windows (our Mac and Linux machines behave exactly as expected) we're experiencing occasional 500 errors while running our Dusk tests, with the logs indicating that somehow the `.env` wasn't loaded for a given request (requests adjacent to this have no such issue). We know it's specifically happening after updating the exe in this merge, and yes, we've doubled checked the php version, chrome version, everything else. These files are definitely messing our Windows dev up. Unfortunately, that's about all we can determine for sure: the specific placement of the 500 error varies, as does the total number of failures (usually two or three spread out across a hundred tests). 

Our best guess is that maybe it's something to do with the session changes mentioned in the https://chromedriver.storage.googleapis.com/2.45/notes.txt? I wouldn't have thought that would be enough to induce a 500, but after two business days of debugging we're at a loss.

driesvints:
@tentus a Chromedriver upgrade shouldn't be breaking because it's just providing you with the latest environment to test on. If you're experiencing issues can you open up an issue so we can track them down and have a look at them?

PR #595, opened 2018-12-27
Update jquery to 3.3.1

staabm:
in case its important - release notes:

https://blog.jquery.com/2018/01/20/jquery-3-3-1-fixed-dependencies-in-release-tag/


driesvints:
Hey @staabm, thanks for sending this in. As per our release notes we don't accept any asset files from pull requests due to security reasons: https://laravel.com/docs/5.7/contributions#compiled-assets

Please also always state why this PR is needed. If there's no reason except for a version bump I don't see the benefit to merge this in.

PR #598, opened 2019-01-01
[5.0] Added waitUntilVue and waitUntilVueIsNot

kapi2289:
## Description
Added 2 new functions in `WaitsForElements` concern
```php
function waitUntilVue($key, $value, $componentSelector = null, $seconds = null)
function waitUntilVueIsNot($key, $value, $componentSelector = null, $seconds = null)
```

## Usage example
```php
public function testExample()
{
        $browser->visit('/')
                ->click('#load')
                ->waitUntilVueIsNot('user.name', null, '@user')
                ->click('#next');
    });
}
```


driesvints:
The `Browser` class is macroable so you can add these methods to your own app if you like: https://laravel.com/docs/5.7/dusk#browser-macros

Issue #530, opened 2018-07-10
Support invocation of dusk testcases through phpdbg


chanaksha:
We use dusk to run our browser as well as all our unit/http test cases for consistency. As dusk accepts all phpunit options and pass them through, it works pretty well.

Not until I tried phpdbg though. For performance reasons I wanted to try running our test suite via phpdbg (instead of XDebug). However it doesn't work

```
$ phpdbg -q -rr artisan dusk
[Welcome to phpdbg, the interactive PHP debugger, v0.5.0]
To get help using phpdbg type "help" and press enter
[Please report bugs to <http://bugs.php.net/report.php>]
[Successful compilation of /Users/shiva/Sandbox/sites/play/vendor/phpunit/phpunit/phpunit]
prompt> run
PHPUnit 7.2.6 by Sebastian Bergmann and contributors.



Time: 24 ms, Memory: 6.00MB

No tests executed!
[Script ended normally]
prompt>
```

Above is just for explanation (no actual tests). The phpdbg goes to interactive prompt even after passing "-q -rr" options.

Looking at DuskCommand.php what I see is that there is no way to pass any arguments to PHP_BINARY when it invokes phpunit. As it doesn't pass "-q -rr" arguments to phpdbg, the test execution enters into phpdbg prompt which I don't want to happen.

We need to provide a way to provide arguments to PHP_BINARY or we need to retain the arguments.

If this is a reasonable request, I am willing to work on a pull request to support it.

I am aware that it wouldn't make sense to run browser test via phpdbg, however this request is such that we can run all test cases using dusk command.

driesvints:
Yeah I think that passing a flag to the PHP binary could be added. Feel free to send in a PR.

chanaksha:
Alright, thanks. All I needed was a nod :) Will try it out in spare time.

PR #602, opened 2019-01-26
Fixes Support invocation of dusk testcases through phpdbg #530

chanaksha:
I have checked how php unit and symphony process handles phpdbg and it looks like hard coding -qrr option as they do is simplest solution. Initially I had thought about generalizing PHP_OPTIONS somehow but php doesn't provide nice way to determine arguments sent to php/phpdbg command itself (and not the script). Other alternative is to rely on environment variables which doesn't look nice either.

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -81,6 +81,10 @@ class DuskCommand extends Command
      */
     protected function binary()
     {
+        if ('phpdbg' === \PHP_SAPI) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Need more information. How do we recreate this?

staudenmeir:
Taylor doesn't read closed PRs, please submit a new one.

chanaksha:
Thanks for the hint :) Couldn't find option to reopen the PR myself.

PR #613, opened 2019-02-23
[5.0] Use custom DISPLAY variable when set on Linux

mbardelmeijer:
Use case: running Dusk on a server environment and using X11 forwarding via SSH to take a peek in the browser running the actions. This is particularity useful in combination with the https://github.com/nunomaduro/laravel-console-dusk package.

We automate specific browser actions on the server side and it's handy to see why it's failing on the server (for instance, we ran into issues because the user agent of the headless browser was different).

Am not aware of any side effects but happy to hear some thoughts about this.

### DIFF
--- a/src/Chrome/ChromeProcess.php
+++ b/src/Chrome/ChromeProcess.php
@@ -79,7 +79,7 @@ class ChromeProcess
             return [];
         }
 
-        return ['DISPLAY' => ':0'];
+        return ['DISPLAY' => env('DISPLAY', ':0')];
     }
 
     /**
--- a/src/Chrome/ChromeProcess.php
+++ b/src/Chrome/ChromeProcess.php
@@ -79,7 +79,7 @@ class ChromeProcess
             return [];
         }
 
-        return ['DISPLAY' => env('DISPLAY', ':0')];
+        return ['DISPLAY' => getenv('DISPLAY') ?: ':0'];
     }
 
     /**


taylorotwell:
Maybe `DUSK_DISPLAY`?

mbardelmeijer:
The benefit of reading the `DISPLAY` variable it this variable is automatically set. For instance, when using Putty with X11 forwarding you get:

```
michel@web0:~# echo $DISPLAY
localhost:10.0
```

In the case it's unset (most Linux servers) it will fallback to `:0`

taylorotwell:
Why not read from $_ENV?

mbardelmeijer:
No specific reason. Reading from `$_ENV`, `$_SERVER` or via `getenv` all gives the same result. Happy to update it to `$_ENV` if you prefer that.

yohanesgultom:
> No specific reason. Reading from `$_ENV`, `$_SERVER` or via `getenv` all gives the same result. Happy to update it to `$_ENV` if you prefer that.

I'm using version 5.5 (this merge included) but ended up have to add 
```
$_ENV['DISPLAY'] = getenv('DISPLAY');
``` 
in my `DuskTestCase` because `$_ENV['DISPLAY']` is empty during execution of `php artisan dusk`. Do you aware of any case that such thing may happen? I'm using Ubuntu 18.04 LTS.

PR #616, opened 2019-02-27
fixes issue 614

### LINKED ISSUE #614: Error after upgrading to Laravel 5.8
After updating to Laravel 5.8, when running php artisan dusk, I get:

Symfony\Component\Debug\Exception\FatalThrowableError  : Argument 1 passed to Dotenv\Dotenv::__construct() must be an instance of Dotenv\Loader, string given, called in /vendor/laravel/dusk/src/Console/DuskCommand.php on line 208

SmoDav:
Fix #614

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -209,6 +209,12 @@ class DuskCommand extends Command
      */
     protected function refreshEnvironment()
     {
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks for sending this in. I've released v5.0.1 which should fix this. I've reversed the behavior of this PR by making the new behavior the default.

PR #620, opened 2019-03-01
Avoid deprecation messages on PHPUnit 8

mnabialek:
`assertContains` and `assertNotContains` are deprecated in PHPUnit 8, so we need to use `assertTrue` and `assertFalse` together with `Str::contains` if we want to also support PHPUnit < 7.5

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -205,8 +205,8 @@ trait MakesAssertions
      */
     public function assertSourceHas($code)
     {
-        PHPUnit::assertContains(
-            $code, $this->driver->getPageSource(),
+        PHPUnit::assertTrue(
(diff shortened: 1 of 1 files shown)


driesvints:
Heya, thanks for sending this in. I've, however, replaced these with the proper PHPUnit assertions instead of using our own `Str` helper. Here's the commit: https://github.com/laravel/dusk/commit/d48f9905b895c9ac252b1a855d843de2d0878b15

mnabialek:
@driesvints But it won't work when someone is using PHPUnit 7, those assertions were introduced in PHPUnit 7,5 I believe and in composer.json we allow any PHPUnit 7 version

driesvints:
@mnabialek good catch 👍 

Wasn't aware that they were only added then.

PR #627, opened 2019-03-12
[6.0] Refactor #625

staudenmeir:
More elegant but breaking version of #625.


taylorotwell:
Why introduce breaking changes needlessly?

Issue #641, opened 2019-04-24
Update Chrome driver to support browser v74


markdwhite:
- Dusk Version: 5.0.3
- Laravel Version: 5.8.14
- PHP Version: 7.2.17
- Database Driver & Version: mySQL 5.7.25

### Description:
Chrome 74 is now available. The current driver has not been updated for 4 months and supports Chrome 70-73 only.

(Driver info: chromedriver=2.45.615279 (12b89733300bd268cff3b78fc76cb8f3a7cc44e5),platform=Linux 4.15.0-47-generic x86_64)
Facebook\WebDriver\Exception\SessionNotCreatedException: session not created: Chrome version must be between 70 and 73

However...chromedriver 74.X does not explicitly support older versions:
https://chromedriver.storage.googleapis.com/74.0.3729.6/notes.txt

Compared to 2.45.X which clearly supports a range of versions 70-73 (and 2.46) but not 74:
https://chromedriver.storage.googleapis.com/2.45/notes.txt
https://chromedriver.storage.googleapis.com/2.46/notes.txt

Does this mean we have to wait for a version that has a more encompassing support so we don't lose backwards compatibility? So we're either pinning Chrome to 73 or using a standalone driver rather than the packaged driver?

Cheers.

Kumidan:
I'm facing the exact same issue.

ahtinurme:
Same here.

tentus:
Same.

rmeira:
same isue

staudenmeir:
In the meantime, you can use my updater package: https://github.com/staudenmeir/dusk-updater

markdwhite:
> In the meantime, you can use my updater package: https://github.com/staudenmeir/dusk-updater

The reply is appreciated and hopefully it will help some as copying the 74.X chromedriver executable into laravel/dusk/bin will work for local dev. However it doesn't work for CI and an update to the package will be needed at some point. Which may well not happened until a 2.47 chromedriver is released that covers a few versions of Chrome.

staudenmeir:
The ChromeDriver release strategy has changed and it looks like there will be no more releases that support multiple versions of Chrome: http://chromedriver.chromium.org/downloads/version-selection

Why doesn't the updater work for CI?

markdwhite:
> The ChromeDriver release strategy has changed and it looks like there will be no more releases that support multiple versions of Chrome

[...]

PR #643, opened 2019-04-26
[5.0] Implement ChromeDriverCommand

driesvints:
This PR implements the functionality offered by https://github.com/staudenmeir/dusk-updater which was written by @staudenmeir.

This command will allow users to update their chrome driver for their OS. It has a flag that will also allow users to update every binary if they want. I've added this to the install command so the driver is updated whenever the install command is run. This means we don't have to update the docs.

I wrote this as an install command so we can remove the binaries in a future version of Dusk and always have them installed with this command. This will save having them to be packaged with the repo. ~It's also written in such a way that it only installs the binary for the current OS.~

Fixes https://github.com/laravel/dusk/issues/641

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
 Copyright (c) Taylor Otwell
+Copyright (c) Jonas Staudenmeir
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,8 @@
     ],
     "require": {
         "php": ">=7.1.0",
+        "ext-json": "*",
+        "ext-zip": "*",
         "facebook/webdriver": "^1.3",
         "nesbot/carbon": "^1.20|^2.0",
         "illuminate/console": "~5.7.0|~5.8.0|~5.9.0",
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -21,16 +21,6 @@ class InstallCommand extends Command
     protected $description = 'Install Dusk into the application';
 
     /**
-     * Create a new command instance.
-     *
-     * @return void
-     */
-    public function __construct()
-    {
-        parent::__construct();
-    }
-
-    /**
      * Execute the console command.
      *
      * @return mixed
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -57,6 +57,10 @@ class InstallCommand extends Command
         }
 
         $this->info('Dusk scaffolding installed successfully.');
+
+        $this->info('Installing ChromeDriver binaries...');
+
+        $this->call('dusk:chrome-driver', ['--all']);
     }
 
     /**
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -60,7 +60,7 @@ class InstallCommand extends Command
 
         $this->info('Installing ChromeDriver binaries...');
 
-        $this->call('dusk:chrome-driver', ['--all']);
+        $this->call('dusk:chrome-driver', ['--all' => true]);
     }
 
     /**
--- a/src/Chrome/ChromeProcess.php
+++ b/src/Chrome/ChromeProcess.php
@@ -3,7 +3,7 @@
 namespace Laravel\Dusk\Chrome;
 
 use RuntimeException;
-use Illuminate\Support\Str;
+use Laravel\Dusk\OperatingSystem;
 use Symfony\Component\Process\Process;
 
 class ChromeProcess
@@ -87,7 +87,7 @@ class ChromeProcess
      */
     protected function onWindows()
     {
-        return PHP_OS === 'WINNT' || Str::contains(php_uname(), 'Microsoft');
+        return OperatingSystem::onWindows();
     }
 
     /**
@@ -97,6 +97,6 @@ class ChromeProcess
      */
     protected function onMac()
     {
-        return PHP_OS === 'Darwin';
+        return OperatingSystem::onMac();
     }
 }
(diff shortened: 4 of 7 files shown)


driesvints:
@staudenmeir I'd greatly appreciate a review from you :)

staudenmeir:
Not a review, but things I've noticed:

- The `resources/legacy.json` file is missing.
- Can we reuse the OS detection from `ChromeProcess` (by moving it to a shared place)?
- Since the code is basically a copy of my package, it should/must include some kind of copyright notice.

composer.json
```diff
@@ -11,6 +11,8 @@
     ],
     "require": {
         "php": ">=7.1.0",
+        "ext-json": "*",
```
staudenmeir: I don't think we need to add `ext-json`, it's already required by Laravel.

src/Console/InstallCommand.php
```diff
@@ -67,6 +57,10 @@ public function handle()
         }
 
         $this->info('Dusk scaffolding installed successfully.');
+
+        $this->info('Installing ChromeDriver binary...');
+
+        $this->call('dusk:chrome-driver');
```
staudenmeir: We should use the `--all` option here. If you install Laravel & Dusk on a Windows/Mac machine and then run the tests in Homestead, you aren't using the latest ChromeDriver.

composer.json
driesvints: We do since we use `json_` methods in this library. It's always better to explicitly state your dependencies in your composer.json, even if it's used in Laravel.

src/Console/InstallCommand.php
driesvints: Hah, I was doubting to do that but you're right. I've updated this.

driesvints:
> The resources/legacy.json file is missing.

I've added this as an array in the command.

> Can we reuse the OS detection from ChromeProcess (by moving it to a shared place)?

Yeah I was thinking that as well. That's probably better indeed. I've pushed a class that re-uses this knowledge.

> Since the code is basically a copy of my package, it should/must include some kind of copyright notice.

You're absolutely right. I've updated the license file.

driesvints:
@staudenmeir updated.

src/Console/InstallCommand.php
```diff
@@ -67,6 +57,10 @@ public function handle()
         }
 
         $this->info('Dusk scaffolding installed successfully.');
+
+        $this->info('Installing ChromeDriver binaries...');
+
+        $this->call('dusk:chrome-driver', ['--all']);
```
staudenmeir:
```suggestion
        $this->call('dusk:chrome-driver', ['--all' => true]);
```

driesvints: Fixed. Good catch.

composer.json
staudenmeir: Now that `legacy.json` is stored in an array, we can actually remove `ext-json`.

driesvints: I noticed some other usages in the library so I'm gonna leave it in.

driesvints:
@staudenmeir thanks for all your work on this btw 👍

staudenmeir:
When I updated my package, I noticed a potential issue with the new ChromeDriver release strategy:
There will now be beta releases for upcoming Chrome versions and `latestChromeVersion()` might incorrectly select them as the latest version (technically correct, but not really useful).

We can't do anything about that at the moment, we just have to wait for the next release to see what the index page will look like.

LICENSE.md
```diff
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
 Copyright (c) Taylor Otwell
+Copyright (c) Jonas Staudenmeir
```
taylorotwell: I don't think it's common to add to this file on contributions to open source projects. Would we add everyone that has ever contributed to Dusk?

staudenmeir: We don't have to change the license file, a copyright notice in `ChromeDriverCommand.php` would be sufficient.

driesvints:
@staudenmeir I've pushed a commit for this, does this look good to you? https://github.com/laravel/dusk/commit/7e5414b488bbf1e59bbfa77bd0a284cc228678f8

LICENSE.md
driesvints: @staudenmeir I've pushed a commit for this, does this look good to you? https://github.com/laravel/dusk/commit/7e5414b488bbf1e59bbfa77bd0a284cc228678f8

staudenmeir: Please also include the package URL.

driesvints: Done!

staudenmeir: Thanks!

bobbybouwmann:
@driesvints I did found some errors when running this command
```
Facebook\WebDriver\Exception\SessionNotCreatedException: session not created: Chrome version must be between 70 and 73
```
So for me it installed the latest Chrome which is 74, as you also can see on the homepage: http://chromedriver.chromium.org/home

However the Facebook driver doesn't support this version just yet, at least that's what the error is telling me. I've been looking into the `facebook\php-webdriver` package but couldn't find where this check is done!

A work around for now was downloading an older version manually, but the command currently doesn't work whenever a new version is released and not updated in the `php-webdriver` package.

driesvints:
Hey @bobbybouwmann, did you read through https://github.com/laravel/dusk/issues/641 to see if any of the answers helps you?

> So for me it installed the latest Chrome which is 74, as you also can see on the homepage

The error says `Chrome version must be between 70 and 73` so did you try to install version 71 or 72? What's your current chrome browser version? Did you try to upgrade your Chrome browser and quit and restart afterwards? Perhaps even reboot your computer to make sure no processes are lingering anymore?

> However the Facebook driver doesn't support this version just yet

I guess I can't help with this and you'll need to use a lower version for now...

staudenmeir:
> However the Facebook driver doesn't support this version just yet, at least that's what the error is telling me. 

Do you mean the error that you posted? This error is not caused by an incompatibility with the `facebook/webdriver` package.

bobbybouwmann:
@driesvints I had indeed 74 installed, but that was because of the `php artisan du`sk:chrome-driver` command. I installed 72 and everything started working again! Just wanted to give a heads up!

PR #645, opened 2019-05-04
[5.1] Adds automatic Chrome version detection

peterfox:
Adds the `—detect` option for `dusk:chrome-driver` that can be provided a custom path or it will use an expected default for the operating system. It will then try to automatically detect the version of Chrome being used and then download the correct chrome driver for that version. Will make it easier for developers to side step having to check which version of Chrome they have as the current default is for the command to simply grab the latest driver version but sometimes this will not line up with what they have installed.

Haven't tested it for Windows or Linux due to a lack of access to those OSs to test on, I can't see there being a problem with Linux but it's hard to tell with Windows and I just don't have a machine to hand to test it with.

### DIFF
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -5,6 +5,7 @@ namespace Laravel\Dusk\Console;
 use ZipArchive;
 use Illuminate\Console\Command;
 use Laravel\Dusk\OperatingSystem;
+use Symfony\Component\Process\Process;
 
 /**
  * @copyright Originally created by Jonas Staudenmeir: https://github.com/staudenmeir/dusk-updater
@@ -16,7 +17,9 @@ class ChromeDriverCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'dusk:chrome-driver {version?} {--all : Install a ChromeDriver binary for every OS}';
+    protected $signature = 'dusk:chrome-driver {version?} '.
+    '{--all : Install a ChromeDriver binary for every OS} '.
+    '{--detect= : Automatically detect version of Google Chrome installed based on OS default or provided path}';
 
     /**
      * The console command description.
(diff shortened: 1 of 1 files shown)


staudenmeir:
Unfortunately, this is more complicated.

`--version` doesn't work on Windows: https://bugs.chromium.org/p/chromium/issues/detail?id=158372

We can use `reg query "HKEY_CURRENT_USER\Software\Google\Chrome\BLBeacon" /v version` instead:

```
HKEY_CURRENT_USER\Software\Google\Chrome\BLBeacon
    version    REG_SZ    74.0.3729.131
```

You can use Homestead to test it on Linux.

We should also support Google Chrome: `/usr/bin/google-chrome`

src/Console/ChromeDriverCommand.php
```diff
@@ -127,15 +141,44 @@ public function handle()
         $this->info(sprintf($message, $all ? 'binaries' : 'binary', $version));
     }
 
+    protected function detectVersion($currentOS, $path = null)
```
staudenmeir: Please add the PHPDoc.

src/Console/ChromeDriverCommand.php
```diff
@@ -127,15 +141,44 @@ public function handle()
...
+    {
+        $installPath = $path ? realpath($path) : $this->installPaths[$currentOS];
+
+        if (! file_exists($installPath) || is_dir($installPath) || ! is_executable($installPath)) {
```
staudenmeir: Do we need the `is_dir()` check? Doesn't `file_exists()` cover that?

taylorotwell:

staudenmeir:
I've added the Chrome version detection to my updater package: [staudenmeir/dusk-updater](https://github.com/staudenmeir/dusk-updater)

browner12:
Can we re-address this and start by making it work for Unix systems?  Seems like we have that part solved.  It seems silly to hold this back because Windows doesn't work.

PR #655, opened 2019-06-17
[5.0] add option to fullsize the browser

browner12:
this will calculate the size of the `<body>` and resize the browser to match it.

this can be especially helpful taking screenshots due to failures, because often times we cannot see all the content in the screenshot, making it difficult to debug.  eventually it may be nice to have an option to automatically screenshot at fullsize, but we'll start with this PR.


taylorotwell:
Renamed to `fitContent`.

PR #656, opened 2019-06-29
Ability to customize attribute name for hooking a selector

slavarazum:
Using static `dusk` attribute for hooking a selector, can compromise which tools we are using on server. It would be better to have ability to customize it!

### DIFF
--- a/src/Dusk.php
+++ b/src/Dusk.php
@@ -7,7 +7,7 @@ use InvalidArgumentException;
 class Dusk
 {
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't think it would actually be better.

slavarazum:
@taylorotwell So, do you think that change a compromising `dusk` keyword is unnecessary? If yes, why do you think so?

P.S. Why you always don't give any feedback on PRs? It's really demotivate to make any contributions.

PR #657, opened 2019-07-01
Comply to PSR-2 coding standards

CaddyDz:
Only one argument is allowed per line in a multi-line function call


driesvints:
Code in the Laravel libraries often structures its arguments this way.

PR #661, opened 2019-07-28
Add assertion to check if a button is disabled

SjorsO:
This PR adds the following two assertions:
```php
public function assertButtonEnabled($button)

public function assertButtonDisabled($button)
```

Asserting that a button is disabled is something I frequently do, I feel like it deserves its own assertion. Both new assertions select a button in the same way `->press($button)` does.

Without these assertions, checking if a button is disabled has to be done something like this:
```php
->tap(function (Browser $browser) {
    $this->assertFalse($browser->resolver->resolveForButtonPress('Press me')->isEnabled());
})
```

There is already a `assertDisabled($field)` assertion, but since that uses `->resolveForField()` it can not be used to select buttons.

### DIFF
--- a/tests/MakesAssertionsTest.php
+++ b/tests/MakesAssertionsTest.php
@@ -137,9 +137,9 @@ class MakesAssertionsTest extends TestCase
         );
         $browser = new Browser($driver, $resolver);
 
-        $browser->assertButtonIsEnabled('Press me');
+        $browser->assertButtonEnabled('Press me');
 
-        $browser->assertButtonIsEnabled('Cant press me');
+        $browser->assertButtonEnabled('Cant press me');
     }
 
     public function test_assert_button_is_disabled()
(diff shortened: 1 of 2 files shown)


staudenmeir:
I think `assertButtonEnabled()` and `assertButtonDisabled()` would be more consistent with the existing assertions.

SjorsO:
Good point, I've renamed them

tests/MakesAssertionsTest.php
```diff
@@ -124,6 +124,42 @@ public function test_assert_disabled()
         }
     }
 
+    public function test_assert_button_is_enabled()
```
staudenmeir:
```suggestion
    public function test_assert_button_enabled()
```

tests/MakesAssertionsTest.php
```diff
@@ -124,6 +124,42 @@ public function test_assert_disabled()
...
+            false
+        );
+        $browser = new Browser($driver, $resolver);
+
+        $browser->assertButtonEnabled('Press me');
+
+        $browser->assertButtonEnabled('Cant press me');
+    }
+
+    public function test_assert_button_is_disabled()
```
staudenmeir:
```suggestion
    public function test_assert_button_disabled()
```

PR #663, opened 2019-08-01
[5.0] Add assertConsoleLogMissingErrors assertion

SjorsO:
This PR adds a `assertConsoleLogMissingErrors()` assertion that asserts that the console log does not contain any unexpected errors.

This assertion is especially useful when testing javascript heavy applications, such as an SPA made with Vue or React. Because javascript errors don't always break the test, they can be hard to discover. Using the `assertConsoleLogMissingErrors()` assertion at the end of your Dusk test gives extra confidence that nothing went wrong.

When the assertion fails, it looks something like this:
image

However, many console messages can be safely ignored. To ignore console log messages, you can define a custom closure. The code below contains an example of the messages we ignore in a legacy react SPA that we started adding Dusk tests to a few months ago (we are already using this assertion as a macro).
```php
Browser::$assertConsoleLogFilter = function ($consoleLog) {
    $ignoreIfMessageContains = [
        'Warning: Failed prop type:',
        'Warning: validateDOMNesting(...): %s cannot appear as a descendant of',
        'Warning: validateDOMNesting(...): %s cannot appear as a child of',
        'Warning: Cannot update during an existing state transition (such as within `render`)',
        'Warning: Each child in a list should have a unique \"key\" prop.',
        'Warning: `value` prop on `%s` should not be null',
        'Warning: A component is changing a controlled input of type %s to be uncontrolled',
    ];

    foreach ($ignoreIfMessageContains as $ignoreText) {
        if (stripos($consoleLog['message'], $ignoreText) !== false) {
            return false;
        }
    }

    return true;
};
```

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -47,6 +47,13 @@ class Browser
     public static $storeConsoleLogAt;
 
     /**
+     * The closure that is used to filter console messages for console log assertions.
+     *
+     * @var Closure
+     */
+    public static $assertConsoleLogFilter;
+
+    /**
      * The browsers that support retrieving logs.
      *
      * @var array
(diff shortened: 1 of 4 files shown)


taylorotwell:
Maybe just macro this into your own browser via a package for now.

PR #664, opened 2019-08-10
Method to wait for text in selector

KaiReichart:
Added Method to wait for text in selector.

```php
public function waitForTextIn($selector, $text, $seconds = null)

```
Example usage:

```php
$this->browse(function(Browser $browser) use($user){
    $browser->loginAs($user)
        ->visit('/datatable')
        ->waitForTextIn('@datatableColStatus', STATUS_DONE, 60);
});
```

### DIFF
--- a/src/Concerns/WaitsForElements.php
+++ b/src/Concerns/WaitsForElements.php
@@ -85,7 +85,7 @@ trait WaitsForElements
             return Str::contains($this->resolver->findOrFail('')->getText(), $text);
         }, $message);
     }
-    
+
     /**
      * @param string $selector
      * @param string $text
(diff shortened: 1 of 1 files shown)


src/Concerns/WaitsForElements.php
```diff
@@ -86,6 +86,22 @@ public function waitForText($text, $seconds = null)
         }, $message);
     }
 
+    /**
+     * @param string $selector
```
staudenmeir: Please format the PHPDoc properly. Take a look at `waitForText()`.

taylorotwell:
Invalid formatting.

src/Concerns/WaitsForElements.php
KaiReichart: Changed it in my Fork, should I start a new PR?

Issue #666, opened 2019-08-21
Please change "dusk" attribute to "data-dusk"


adrienne:
Currently the documentation says the following:

> Dusk selectors allow you to focus on writing effective tests rather than remembering CSS selectors. To define a selector, add a `dusk` attribute to your HTML element. Then, prefix the selector with `@` to manipulate the attached element within a Dusk test

This creates invalid HTML. Can you please change it (or at least allow a configuration option) so that instead of `dusk` the relevant attribute is `data-dusk`? All custom attributes in HTML are supposed to be prefixed with `data-`.

driesvints:
Feel free to send in a pr

adrienne:
@driesvints - I was just annoyed enough to do this. Here. https://github.com/laravel/dusk/pull/669

pataar:
FWIW: https://github.com/laravel/dusk/pull/1012

rossjcooper:
For anyone who comes here from Google you can now do this and it is documented here https://laravel.com/docs/10.x/dusk#dusk-selectors

dualklip:
> For anyone who comes here from Google you can now do this and it is documented here https://laravel.com/docs/10.x/dusk#dusk-selectors

Thank you @rossjcooper . That what exactly I needed

PR #669, opened 2019-09-09
Configurable selector prefix (related to issue #666)

adrienne:
This fairly simple change allows users to configure the `dusk=""` prefix that Dusk uses to parse selectors with the `@` prefix. This is intended primarily to allow users to namespace the prefix into `data-[whatever]`, for compatibility with HTML validators. Validators are still important to many people's workflows. The default is still `dusk=""`, to avoid breakage of existing code.

There is *no new test coverage*, because the test `test_format_correctly_formats_selectors` in the file `https://github.com/laravel/dusk/blob/5.0/tests/ElementResolverTest.php` covers this change.

### DIFF
--- a/config/dusk.php
+++ b/config/dusk.php
@@ -1,0 +1,16 @@
+    
+<?php
+
+return [
+    /*
+    |--------------------------------------------------------------------------
+    | Dusk Selector Prefix
+    |--------------------------------------------------------------------------
+    |
+    | This is the prefix Dusk will use when resolving selectors prefixed with @
+    | (see https://laravel.com/docs/master/dusk#dusk-selectors)
+    |
+    */
+    'selector_prefix' => 'dusk',
+
+];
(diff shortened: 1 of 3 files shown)


SjorsO:
Maybe instead of adding a new config file, we can add an instance variable on the `BaseTestCase`. Then we can override that in our `DuskTestCase`, and not have to publish a whole config file for a small change.

---

I like PR. It also prevents my IDE from complaining in `.tsx` files about the invalid attribute:
image

With this PR, you can use this instead:
image

driesvints:
This seems to break both the build and StyleCI. Feel free to resubmit with passing tests and code style corrections.

Rikaelus:
For reference for those who find their way here, I implemented similar functionality based on a `DUSK_HTML_ATTRIBUTE` environment variable in a [now-merged 7.x PR](https://github.com/laravel/dusk/pull/1022). Hopefully the slightly different approach will keep it on the happy side of tests.

Issue #672, opened 2019-09-11
Chrome 77 dusk:install broken with chrome 77


lionslair:
- Dusk Version: 5.5.8
- Laravel Version: 5.8.35
- PHP Version: 7.3.8
- Database Driver & Version: mysql 5.7.27

### Description:
The dusk install command no longer works with the latest version of chrome  77

### Steps To Reproduce:

php artisan dusk:install
Dusk scaffolding installed successfully.
Downloading ChromeDriver binaries...
In ChromeDriverCommand.php line 172:
                       
  Undefined offset: 1

ahmadsyamim:
same here..

php artisan dusk:chrome-driver

   ErrorException  : Undefined offset: 1

lionslair:
They have changed the HTML around the string 


`<li><span style="background-color:transparent">Latest stable&nbsp;</span>release: <a href="https://chromedriver.storage.googleapis.com/index.html?path=77.0.3865.40/" style="background-color:transparent" target="_blank">ChromeDriver 77.0.3865.40</a></li>

<li>Latest stable release: <a href="https://web.archive.org/web/20190705211833/https://chromedriver.storage.googleapis.com/index.html?path=75.0.3770.90/" style="background-color:transparent">ChromeDriver 75.0.3770.90</a></li>`

it is now 

`<span style="background-color:transparent">Latest stable&nbsp;</span>release:`

not

`Latest stable release:`

browner12:
fixed by #674

lionslair:
When will the release occur?

mattdfloyd:
How about grabbing the release number from https://chromedriver.storage.googleapis.com/LATEST_RELEASE ?

staudenmeir:
@mattdfloyd Until yesterday, this file had been deprecated: https://bugs.chromium.org/p/chromedriver/issues/detail?id=3128

PR #673, opened 2019-09-11
Fix pattern for fetch last release of chromedriver

MilesChou:
Release element is change so that regex cannot fetch the version now.

The release element today is:

```
<span style="background-color:transparent">Latest stable </span>release:  <a href="https://chromedriver.storage.googleapis.com/index.html?path=77.0.3865.40/"
```

So we can use this regex to fetch version:

```
'/release:.*?\?path=([\d.]+)/'
```

### DIFF
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -167,7 +167,7 @@ class ChromeDriverCommand extends Command
     {
         $home = $this->getUrl($this->homeUrl);
 
-        preg_match('/Latest stable release:.*?\?path=([\d.]+)/', $home, $matches);
+        preg_match('/release:.*?\?path=([\d.]+)/', $home, $matches);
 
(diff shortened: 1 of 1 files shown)


lionslair:
I made an issue but you just beat me to the fix. 
https://github.com/laravel/dusk/issues/672

linuxjuggler:
I think using `Latest stable` instead of `release` would be much accurate to make sure that we only get the results for the `latest stable` incase they change anything in the future.

corneyl:
Wouldn't it be better to not rely on possibly changing scraped information and just install the same chrome-driver version as the installed chrome version? Like this command does e.g.:
```
php artisan dusk:chrome-driver $(google-chrome --version | sed -E "s/[a-zA-Z ]*([0-9]+).*/\\1/")
```

taylorotwell:
Already merged by another PR.

PR #676, opened 2019-09-11
WIP: Configurable selector prefix (related to issue #666)

adrienne:
This fairly simple change allows users to configure the `dusk=""` prefix that Dusk uses to parse selectors with the `@` prefix. This is intended primarily to allow users to namespace the prefix into `data-[whatever]`, for compatibility with HTML validators. Validators are still important to many people's workflows. The default is still `dusk=""`, to avoid breakage of existing code.

There is *no new test coverage*, because the test `test_format_correctly_formats_selectors` in the file `https://github.com/laravel/dusk/blob/5.0/tests/ElementResolverTest.php` covers this change.

### DIFF
--- a/config/dusk.php
+++ b/config/dusk.php
@@ -1,0 +1,16 @@
+    
+<?php
+
+return [
+    /*
+    |--------------------------------------------------------------------------
+    | Dusk Selector Prefix
+    |--------------------------------------------------------------------------
+    |
+    | This is the prefix Dusk will use when resolving selectors prefixed with @
+    | (see https://laravel.com/docs/master/dusk#dusk-selectors)
+    |
+    */
+    'selector_prefix' => 'dusk',
+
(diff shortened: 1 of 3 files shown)


adrienne:
@driesvints - Okay, I fixed the style issues but I don't actually know how to fix the test cases. The feature *works* when I run `php artisan dusk`, but there's something up with the `config()` helper function -- it isn't in the test harness, somehow, so `ElementResolver.php` fails. I tried switching it to the `\Config::` facade but that didn't work either. Please help?

SjorsO:
I think an instance variable on `BaseTestCase` is a better idea than a whole new config file.

driesvints:
Build seems to be failing. Feel free to resend. Also: if you send in a WIP pr you can use the native Github functionality for that.

PR #677, opened 2019-09-16
[5.5] Improve detection of latest stable ChromeDriver release

staudenmeir:
By default, `php artisan dusk:chrome-driver` downloads the latest stable ChromeDriver release. We are currently parsing http://chromedriver.chromium.org to get the version number, but that's not ideal. When the HTML changes (as it did last week), it breaks the updater (#672).

https://chromedriver.storage.googleapis.com/LATEST_RELEASE does contain this version number, but the file had been deprecated for the last months. I explained our issue to the ChromeDriver team and they removed the deprecation: https://bugs.chromium.org/p/chromedriver/issues/detail?id=3128

### DIFF
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -29,14 +29,14 @@ class ChromeDriverCommand extends Command
     protected $description = 'Install the ChromeDriver binary';
 
     /**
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks. I'll do a release for this tomorrow.

PR #682, opened 2019-10-13
[5.x] Add graceful handler for SIGINT for .env restoration

mbardelmeijer:
This PR ensures that if the `php artisan dusk` command is cancelled by a CTRL+C (command cancel, i.e. SIGINT), the original .env is restored correctly.

We often run a dusk test, and cancel within the test suite if an issue comes up. In that case, the `.env.backup` isn't restored, and requires it to be restored manually.

Also adds a `try {} finally {}` clause if the Dusk commands fails themself the environment is still restored.

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -6,8 +6,8 @@ use Dotenv\Dotenv;
 use Illuminate\Console\Command;
 use Illuminate\Support\Str;
 use Symfony\Component\Finder\Finder;
-use Symfony\Component\Process\Exception\RuntimeException;
 use Symfony\Component\Process\Exception\ProcessSignaledException;
+use Symfony\Component\Process\Exception\RuntimeException;
 use Symfony\Component\Process\Process;
 
 class DuskCommand extends Command
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -288,8 +288,8 @@ class DuskCommand extends Command
      */
     protected function removeConfiguration()
     {
-        if (! $this->hasPhpUnitConfiguration) {
-            unlink(base_path('phpunit.dusk.xml'));
+        if (! $this->hasPhpUnitConfiguration && file_exists($file = base_path('phpunit.dusk.xml'))) {
+            unlink($file);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Console/DuskCommand.php
```diff
@@ -68,9 +69,15 @@ public function handle()
...
 
-            return $process->run(function ($type, $line) {
-                $this->output->write($line);
-            });
+            try {
+                return $process->run(function ($type, $line) {
+                    $this->output->write($line);
+                });
+            } catch (ProcessSignaledException $e) {
+                if ($e->getSignal() !== 2) {
```
driesvints: Can we replace the `2` by a predefined constant?

src/Console/DuskCommand.php
```diff
@@ -169,15 +192,37 @@ protected function withDuskEnvironment($callback)
...
+    /**
+     * Setup the SIGINT signal handler for CTRL+C exits.
+     *
+     * @return void
+     */
+    protected function setupSignalHandler()
+    {
+        declare(ticks=1);
+
+        pcntl_signal(SIGINT, function () {
```
driesvints: `ext-pcntl` will need to be required in the `composer.json` file.

src/Console/DuskCommand.php
mbardelmeijer: Good one, have replaced it with `SIGINT`, global constant provided by the PCNTL extension: https://www.php.net/manual/en/pcntl.constants.php

src/Console/DuskCommand.php
mbardelmeijer: Added. Perhaps better to add it to a `suggest` like the illuminate/queue package does? What do you think is better in this case? See https://github.com/illuminate/queue/blob/master/composer.json#L41

Would prefer it as a suggestion, but am not sure what a good suggestion text would be.

driesvints: It can't be a suggest because it's always executed when running `php artisan dusk`.

mbardelmeijer: True, but if we moved it to suggest if would be together with a `extension_loaded`/`function_exists` call to ensure the extension is loaded there.

But think as Dusk runs only on local machines, the requirement for this extension is minimal.

taylorotwell:
Don't use declare(ticks=1)... use asnyc signals which are available on all versions of PHP that Dusk requires.

taylorotwell:
https://www.php.net/manual/en/function.pcntl-async-signals.php

PR #684, opened 2019-10-23
Add scrollTo method

mbardelmeijer:
Re-submission of https://github.com/laravel/dusk/pull/367.

When using click on elements which are not in the viewport, it's quite handy to add a simple `scrollTo` call to ensure the browser has the element in the viewport. We macro this in all our projects, but would be nicer to have support for this in the core.

Example:
```
$browser->visit(route('overview'))
   ->scrollTo('@create')
   ->click('@create')
```

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -250,6 +250,19 @@ class Browser
     }
 
     /**
+     * Scroll screen to element at the given selector.
+     *
+     * @param  string  $selector
+     * @return $this
+     */
+    public function scrollTo($selector)
+    {
+        $this->driver->executeScript("document.querySelector('$selector').scrollIntoView();");
+
+        return $this;
+    }
+
+    /**
      * Take a screenshot and store it with the given name.
      *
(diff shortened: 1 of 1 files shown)


src/Browser.php
```diff
@@ -249,6 +249,21 @@ public function move($x, $y)
...
+     * Scroll screen to element at the given selector.
+     *
+     * @param  string  $selector
+     * @return $this
+     */
+    public function scrollTo($selector)
+    {
+        $selector = $this->resolver->format($selector);
+
+        $this->driver->executeScript("$(\"html, body\").animate({scrollTop: $(\"$selector\").offset().top}, 0);");
```
crynobone: Does `jQuery` or `$` always present?

driesvints: @crynobone hmm good remark I'll send in a PR for this.

driesvints: https://github.com/laravel/dusk/pull/686

Issue #689, opened 2019-11-06
Issue with ext-pcntl


lk77:
- Dusk Version:5.6.0
- Laravel Version: 6.4.1
- PHP Version: 7.3.1
- Database Driver & Version: mysql 5.7

### Description:

A dependency has been added in 5.6.0 to ext-pcntl, and it's breaking everything, the extension needs to be added to the docker test image. Why such a change on a minor update ? 
This type of changes should go to next major version 6.0 i think.

thanks.

leecrosdale:
@lk77 was there a fix you found for this?

lk77:
@leecrosdale i reverted to 5.5 for now due to another issue, 
but you need to enable pcntl in php, for example for docker :
```
RUN docker-php-ext-install pcntl
```

leecrosdale:
Ah ok, a couple of devs here are running windows, where PCNTL isn't available. 

We ended up doing

`composer install --ignore-platform-reqs`

which seems to work for now!

lk77:
yes, we had the same issue for windows.
Honestly i think this should be an optional dependency.

driesvints:
Hmm, that indeed seems to have broken Windows support. We'll try to reverse it later today.

driesvints:
This was the PR btw: https://github.com/laravel/dusk/pull/682

driesvints:
I send in a PR here: https://github.com/laravel/dusk/pull/692

If anyone can verify that this works on Windows that'd be great.

asbiin:
@driesvints I can confirm it "works" on Windows == I can install it without `--ignore-platform-reqs`, and it can run tests.

PR #692, opened 2019-11-12
[5.x] Prevent using pcntl when not installed

driesvints:
This prevents using pcntl when it's not installed. This was preventing Dusk from being run on Windows.

I have no way of testing this on Windows so I hope this will work.

Fixes https://github.com/laravel/dusk/issues/689


driesvints:
Seems to work on Windows again now: https://github.com/laravel/dusk/issues/689#issuecomment-552831955

PR #693, opened 2019-11-13
Escape single quotes

### LINKED ISSUE #312: Fix bug : Error insert textbox value with single quote
Hai bro developers...:)
I got this error message : Facebook\WebDriver\Exception\UnknownServerException: unknown error: Runtime.evaluate threw exception: SyntaxError: Unexpected identifier
Issue appear when my faker provide **single quote in textbox value.**
**I changed this line to fix it:** https://github.com/laravel/dusk/blob/1.0/src/Concerns/InteractsWithElements.php#L97
from : `"document.querySelector('{$selector}').value = '{$value}';"`
into this one : `"document.querySelector('{$selector}').value = \"{$value}\";"`

Can you add this to next release ? thank you :)

JunaidQadirB:
***Note:*** This might be a rare use-case but it does break when someone passes single quote in the string.
 
The `value` function does not escape the contents of  `$value` variable as a result if someone passes a string even with a escaped single quote, the code breaks because value is assigned using single quotes  on line 68 argument passed to `executeScript`,
`"document.querySelector('{$selector}').value = '{$value}';"`

```php
$browser->value('myField', "Field\'s Name");
```

***Current output of the script:***
```js
document.querySelector('myField').value = 'Field's Name';
```
***Expected output of the script:***
```js
document.querySelector('{$selector}').value = 'Field\'s Name';"
```

Fixes #312

### DIFF
--- a/src/Concerns/InteractsWithElements.php
+++ b/src/Concerns/InteractsWithElements.php
@@ -62,7 +62,9 @@ trait InteractsWithElements
         if (is_null($value)) {
             return $this->resolver->findOrFail($selector)->getAttribute('value');
         }
+
         $value = addslashes($value);
+
         $selector = $this->resolver->format($selector);
 
         $this->driver->executeScript(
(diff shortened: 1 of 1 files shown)


DasRed:
Note: This fix will work for single quotes, but does not work for line breaks.
This solution would be better
```js
$this->driver->executeScript("document.querySelector('{$selector}').value = " . json_encode($value) . ";");
```

DasRed:
I have created a new pull request: https://github.com/laravel/dusk/pull/734

PR #705, opened 2019-12-14
[5.0] store the source of the browser

browner12:
I can't tell you how many times I've clicked on the screenshot of a failed test trying to interact with it...

This will allow us to dig through the full source code of the browser at the point in time of the failures, which can help with debugging.

---

2 questions to ask:

- Should we store the source only on failures or for all tests? Currently I'm storing it for **all** tests.
- What extension should we use for the file? For now I picked `.txt`, but I'm considering `.html`.

---

My grand master plan is to maybe proxy this code somehow so people can pull up the failure in their browser and ***actually*** interact with it.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -47,6 +47,13 @@ class Browser
     public static $storeConsoleLogAt;
 
     /**
+     * The directory that will contain any source files.
+     *
+     * @var string
+     */
+    public static $storeSourceAt;
+
+    /**
      * The browsers that support retrieving logs.
      *
      * @var array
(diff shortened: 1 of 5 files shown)


SjorsO:
Does this also work for SPAs? Or does this only work for "normal" blade/html pages?

taylorotwell:
I don't think we would want to the source code every test? Seems like a lot of wasted space?

browner12:
@SjorsO I don't see why it wouldn't. I guess I'm not *positive* how the driver gets the source code, whether it's the source at the load of the page or if it takes into account all the changes made during the execution of the page.

@taylorotwell I can move this into the `catch` blocks if you'd like.  One of the issues I'm running into right now is if the code is in the `catch` block, it is only executed on **Errors** and not on **Failures**.  If we could figure out a way to execute on Failures, that would be best.

browner12:
do you guys handle new merge conflicts that pop up, or should I go in and do that?

driesvints:
@browner12 best if you rebase yourself

taylorotwell:
Personally I think I would make this a method on the browser that immediately takes a snapshot of the source and stores it. Then, if you have a test that fails you can just pop this method onto the browser somewhere and grab the source. Then we don't have to worry about storing the source on every test or every failure, etc.

You would just end up using when you need to similar to "dd" etc.

PR #706, opened 2019-12-18
[5.x] add "waitUntilMissingText"

SjorsO:
I'm using `waitForText` to ensure a modal has opened. But after closing the modal, there is no opposite method to wait until the modal has disappeared.

### DIFF
--- a/src/Concerns/WaitsForElements.php
+++ b/src/Concerns/WaitsForElements.php
@@ -71,6 +71,25 @@ trait WaitsForElements
     }
 
     /**
+     * Wait for the given text to be removed.
+     *
+     * @param  string  $text
+     * @param  int  $seconds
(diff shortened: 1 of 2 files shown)


src/Concerns/WaitsForElements.php
```diff
@@ -70,6 +70,25 @@ public function waitUntilMissing($selector, $seconds = null)
         }, $message);
     }
 
+    /**
+     * Wait for the given text to be removed.
+     *
+     * @param  string  $text
+     * @param  int  $seconds
+     * @return $this
+     * @throws \Facebook\WebDriver\Exception\TimeOutException
```
driesvints: Please make sure to always add a whiteline before the `@throws` annotation.

PR #717, opened 2020-01-14
[5.0] update `fitContent()`

browner12:
if the user is in an iframe when a failure occurs, `fitContent()` will grab the dimensions of the *iframe* body, not the *root* body that we actually want.

this change makes sure we switch back to the default scope before resizing the browser to fit the content.


browner12:
fixes #715

PR #721, opened 2020-01-18
Update jQuery file

asbiin:
Update jQuery file using the `min` version [here](https://code.jquery.com/jquery-3.4.1.min.js).

The current jQuery file is the `slim` build: this version does not include the `animate` function required to use [`scrollTo`](https://github.com/laravel/dusk/blob/5.0/src/Browser.php#L328) function.


browner12:
similar to how we only allow Laravel employees to update the Chrome Drivers, can we restrict updating this asset to employees as well? It would be too easy for a malicious actor to sneak some code into this due to the size of the change.

Or at a minimum have the employee comment they have confirmed the asset is unaltered from the source, prior to merging?

FWIW, I compared this with the Jquery download page and it looks 👍 .

https://code.jquery.com/jquery-3.4.1.min.js

PR #724, opened 2020-01-26
[5.x] Add ProvidesBrowser::getBrowserCallerName();

crynobone:
This would allow better support when we need to use `laravel/dusk` outside of `PHPUnit\Framework\TestCase`.

`getName(false)` is an internal `PHPUnit\Framework\TestCase` method.


crynobone:
Not sure 5.x is the right branch for this since it would be a small breaking changes to for example `orchestra/testbench-dusk` but I have no idea what's the release plan for 6.0.

PR #734, opened 2020-02-12
[6.x] Bugfix quoting for InteractsWithElements::value

DasRed:
This handles JS quoting a little bit better.

### DIFF
--- a/src/Concerns/InteractsWithElements.php
+++ b/src/Concerns/InteractsWithElements.php
@@ -63,12 +63,10 @@ trait InteractsWithElements
             return $this->resolver->findOrFail($selector)->getAttribute('value');
         }
 
-        $value = addslashes($value);
-
         $selector = $this->resolver->format($selector);
 
         $this->driver->executeScript(
-            "document.querySelector('{$selector}').value = '{$value}';"
+            "document.querySelector(" . json_encode($selector) . ").value = " . json_encode($value) . ";"
         );
 
         return $this;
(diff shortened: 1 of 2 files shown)


driesvints:
Can you please give a bit more info here why this is needed? Some code examples?

Is this a breaking change? If not it could go to the current release branch (5.0).

DasRed:
> Can you please give a bit more info here why this is needed? Some code examples?
> 
> Is this a breaking change? If not it could go to the current release branch (5.0).

Yes of course. 

We found the bug for quoting of JS by calling the method InteractsWithElements::value.
If you call this method with  following arguments
- '#selector'
- 'abc\'def'
will break the generated JS code.

Then we found this pull request: https://github.com/laravel/dusk/pull/693 
But the pull request does not fully fix this problem. If you us an value, which has an line break, then the generated JS Code will break again.

This pull request fix this completely.
No breaking changes

driesvints:
@DasRed tests are failing so that would indicate that there are breaking changes.

DasRed:
> @DasRed tests are failing so that would indicate that there are breaking changes.

I have written the test, which are failing. So no breaking changes.
The tests should run now. I used a new version of PHPUnit, which is not supported by your composer. So i fixed the test, so that they work with your version.

driesvints:
Honestly, can we perhaps have some more code examples at what you're trying to do? Some tests and the piece of html/js you're doing the assertions against?

DasRed:
Okay,

The value function does not correctly escape the contents of $value variable as a result if someone passes a string even with a escaped single quote, the code breaks because value is assigned using single quotes on line 69 argument passed to executeScript,

$browser->value('my\'Field', "long text\nwith line breaks");
Current output of the script:
```js
document.querySelector('my\'Field').value = 'Field
s Name';
```
Expected output of the script:
```js
document.querySelector("my'Field").value ="long text\nwith line breaks";
```

PR #735, opened 2020-02-13
[5.x] Bugfix quoting for InteractsWithElements::value

DasRed:
This handles JS quoting a little bit better.

We found the bug for quoting of JS by calling the method InteractsWithElements::value.
If you call this method with  following arguments
- '#selector'
- 'abc\'def'
will break the generated JS code.

Then we found this pull request: https://github.com/laravel/dusk/pull/693 
But the pull request does not fully fix this problem. If you us an value, which has an line break, then the generated JS Code will break again.

This pull request fix this completely.
No breaking changes

The value function does not correctly escape the contents of $value variable as a result if someone passes a string even with a escaped single quote, the code breaks because value is assigned using single quotes on line 69 argument passed to executeScript,

$browser->value('my\'Field', "long text\nwith line breaks");
Current output of the script:
```js
document.querySelector('my\'Field').value = 'Field
s Name';
```
Expected output of the script:
```js
document.querySelector("my'Field").value ="long text\nwith line breaks";
```

Please take also a closer look at https://github.com/laravel/dusk/pull/734

### DIFF
--- a/src/Concerns/InteractsWithElements.php
+++ b/src/Concerns/InteractsWithElements.php
@@ -66,7 +66,7 @@ trait InteractsWithElements
         $selector = $this->resolver->format($selector);
 
         $this->driver->executeScript(
-            "document.querySelector(" . json_encode($selector) . ").value = " . json_encode($value) . ";"
+            'document.querySelector('.json_encode($selector).').value = '.json_encode($value).';'
         );
 
         return $this;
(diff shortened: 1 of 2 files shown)


driesvints:
Heya thanks. How is this different from #734?

DasRed:
> Heya thanks. How is this different from #734?

Hey, there is no difference between this PR and #734. I made this PR, because we need the fix into v5.x for our testing ;)

A new dusk release would be nice with this PR in next days :)

driesvints:
@DasRed releases are done weekly on Tuesdays so if this gets merged it'll be in the release of next Tuesday.

PR #739, opened 2020-02-21
[5.x] Switch to Github Actions

driesvints:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,62 @@
...
+            phpunit: ^7.5
+          - php: 7.1
+            laravel: ^7.0
+            phpunit: ^8.0
+
+    name: P${{ matrix.php }} - L${{ matrix.laravel }} - U${{ matrix.phpunit }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v1
```
browner12: is there a reason you're not using v2 here?

driesvints: Fixed thanks

PR #759, opened 2020-04-10
[5.x] PHPUnit 9 support

dmitryuk:
PHPUnit >= 9.1 deprecated the method assertRegExp instead of new one assertMatchesRegularExpression.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,17 @@ jobs:
       matrix:
         php: [7.1, 7.2, 7.3, 7.4]
         laravel: [5.7.*, 5.8.*, ^6.0, ^7.0]
-        phpunit: [^7.5, ^8.0, ^9.0]
+        phpunit: [^7.5, ^8.0]
+        include:
+          - php: 7.2.29
+            laravel: ^7.0
+            phpunit: ^9.0
+          - php: 7.3
+            laravel: ^7.0
+            phpunit: ^9.0
+          - php: 7.4
+            laravel: ^7.0
+            phpunit: ^9.0
         exclude:
           - php: 7.1
             laravel: 5.7.*
@@ -35,9 +45,6 @@ jobs:
           - php: 7.1
             laravel: ^7.0
             phpunit: ^8.0
-          - php: 7.1
-            laravel: ^7.0
-            phpunit: ^9.0
 
     name: P${{ matrix.php }} - L${{ matrix.laravel }} - U${{ matrix.phpunit }}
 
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -16,6 +16,16 @@ jobs:
...
+        include:
+          - php: 7.2.29
+            laravel: ^7.0
+            phpunit: ^9.0
+          - php: 7.3
+            laravel: ^7.0
+            phpunit: ^9.0
+          - php: 7.4
+            laravel: ^7.0
+            phpunit: ^9.0
```
driesvints, line 28: Are these necessary?

driesvints, line 28: Instead of including these, you'll need to exclude the incompatible PHP versions and PHPUnit 9 versios below in the exclude list.

src/Concerns/MakesUrlAssertions.php
```diff
@@ -4,7 +4,12 @@
 
 use Illuminate\Support\Arr;
 use PHPUnit\Framework\Assert as PHPUnit;
+use PHPUnit\Framework\Constraint\RegularExpression;
 
+/**
+ * Trait MakesUrlAssertions.
+ * @property \Facebook\WebDriver\Remote\RemoteWebDriver $driver
+ */
```
driesvints, line 12: Remove this DocBlock.

driesvints:
Hmm, `assertMatchesRegularExpression` won't work since it's only recently implemented.

dmitryuk:
> I think we should use `assertMatchesRegularExpression` as an alternative for `assertRegExp` as stated in the changelog: https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-9.1.md#added
It is impossible, because `assertMatchesRegularExpression` has implemented in phpUnit 9.1

driesvints:
Wondering if it's maybe better to just drop Laravel 5.x, PHP 7.1 and PHPUnit 7 in a new major release instead 🤔

dmitryuk:
> Wondering if it's maybe better to just drop Laravel 5.x, PHP 7.1 and PHPUnit 7 in a new major release instead

Yes, it will be great, but currently to support PHPUnit 9 it is enough

taylorotwell:
Apparently this isn't ready to go in its current form so closing for now.

driesvints:
@dmitryuk I'll pick this up from where you left later on. Thanks 👍

driesvints:
This will be available in tomorrow's v6 release.

PR #766, opened 2020-04-21
[6.x] Support scrolling an element into view

### LINKED ISSUE #765: Scrolling an element into view
I realise the [`scrollTo($selector)`](https://github.com/laravel/dusk/blob/6.x/src/Browser.php#L328-L337) method was added to the core of Dusk.

However, it might be useful to make use of the [`Element.scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) JavaScript API which ensures the selected element is actually in view? [Browser support](https://caniuse.com/#feat=scrollintoview) is also good for this API as well as support for [`document.querySelector`](https://caniuse.com/#feat=queryselector).

```php
/**
 * Scroll element into view at the given selector.
 *
 * @param  string  $selector
 * @return $this
 */
public function scrollIntoView($selector)
{
    $selector = addslashes($this->resolver->format($selector));

    $this->driver->executeScript("document.querySelector(\"$selector\").scrollIntoView();");

    return $this;
}
```

u01jmg3:
Support scrolling an element into view

Closes #765

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -320,6 +320,21 @@ class Browser
     }
 
     /**
+     * Scroll element into view at the given selector.
+     *
+     * @param  string  $selector
+     * @return $this
+     */
+    public function scrollIntoView($selector)
+    {
+        $selector = addslashes($this->resolver->format($selector));
+
+        $this->driver->executeScript("document.querySelector(\"$selector\").scrollIntoView();");
+
+        return $this;
+    }
+
+    /**
      * Scroll screen to element at the given selector.
      *
      * @param  string  $selector


taylorotwell:
Is there anything actually broken with `scrollTo`?

u01jmg3:
No, nothing is broken by using jQuery's `scrollTo()`. However, using `scrollIntoView()` is a consistent, well supported method of scrolling to an element and ensuring it is in the current viewport. This is important because it means the element is clickable for further interactions with Dusk. It would also reduce the need for jQuery if that's ever something you wanted to consider.

For reference, an article from 2014 although still relevant: https://luxiyalu.com/difference-between-scrolls/

taylorotwell:
So, if it is going to do the same thing as `scrollTo` is there some reason we wouldn't just change `scrollTo` instead of making a new method?

u01jmg3:
Nope other than `scrollIntoView()` is a more accurate method name as it links with the JavaScript API in use which developers are likely to recognise. However, this is not essential.

PR #771, opened 2020-05-06
[6.x] Uses Pest if available

nunomaduro:
In the future, if people install **Pest** in their Laravel applications, they most likely want to use it in their dusk tests too.

This pull request magically uses Pest if available, otherwise, it uses the default PHPUnit command.

Let me know if you prefer something more customizable like a config or something.

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -88,11 +88,15 @@ class DuskCommand extends Command
      */
     protected function binary()
     {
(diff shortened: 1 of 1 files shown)


src/Console/DuskCommand.php
```diff
@@ -88,11 +88,15 @@ public function handle()
      */
     protected function binary()
     {
+        $command = class_exists(\Pest\Laravel\PestServiceProvider::class)
```
driesvints, line 91: @nunomaduro Can you send a pr that imports this class instead?

PR #772, opened 2020-05-12
[6.x] Add dusk route prefix

dmitryuk:
The PR allows to set route prefix using `LARAVEL_DUSK_PATH` env parameter instead of hardcoding one as `_dusk`. 
I was inspired of that because our project is SPA application, so most requests should go to an `/api` prefix.

### DIFF
--- a/src/DuskServiceProvider.php
+++ b/src/DuskServiceProvider.php
@@ -15,20 +15,25 @@ class DuskServiceProvider extends ServiceProvider
     public function boot()
     {
         if (! $this->app->environment('production')) {
-            Route::get('/_dusk/login/{userId}/{guard?}', [
-                'middleware' => 'web',
-                'uses' => 'Laravel\Dusk\Http\Controllers\UserController@login',
-            ]);
+            Route::group(['prefix' => env('LARAVEL_DUSK_PATH', '_dusk')], function () {
+                Route::get('/login/{userId}/{guard?}', [
+                    'middleware' => 'web',
+                    'uses' => 'Laravel\Dusk\Http\Controllers\UserController@login',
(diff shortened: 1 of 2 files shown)


src/DuskServiceProvider.php
```diff
@@ -15,20 +15,25 @@ class DuskServiceProvider extends ServiceProvider
     public function boot()
     {
         if (! $this->app->environment('production')) {
-            Route::get('/_dusk/login/{userId}/{guard?}', [
-                'middleware' => 'web',
-                'uses' => 'Laravel\Dusk\Http\Controllers\UserController@login',
-            ]);
+            Route::group(['prefix' => env('LARAVEL_DUSK_PATH', '_dusk')], function () {
```
driesvints, line 18: This should be a config value.

driesvints:
I don't really understand why you need this. These are internal Dusk calls and have nothing to do with your app. You shouldn't have the need to overwrite or control these.

src/DuskServiceProvider.php
dmitryuk: I has thought about that at the beginning, but currently where is no config file in the project. So, are you sure that I need to create config file with only 1 key?

dmitryuk:
In short, our architecture contains 2 containers - frontend and backend.
To prevent cors over requests, nginx of frontend container contains proxy_pass to backend with location `/api`. Dusk can't work so.
Idea has taken from an article
https://blog.bitsrc.io/how-and-why-you-should-avoid-cors-in-single-page-apps-db25452ad2f8

taylorotwell:
We can't call the env() function outside of configuration values.

Issue #393, opened 2017-10-20
Change the location of screenshots and console folders


vpratfr:
The tests folder is PSR-4 compliant. Those two folders do not really belong here.

They would rather be located in `storage/logs/dusk/...`

Moving the folders is currently possible with a custom `setUp` method in the DuskTestCase class:

    protected function setUp()
    {
        parent::setUp();

        Browser::$storeScreenshotsAt = storage_path('logs/dusk/screenshots');
        Browser::$storeConsoleLogAt = storage_path('logs/dusk/console');
    }

However, the DuskCommand complains that the folders in tests/Browser do not exist when trying to clean them.

I think that for the same reasons some folders/files were taken out of the `app` folder from L4 to L5, we should keep the tests folder free of non-PSR-4 stuff.

deleugpn:
I don't think there's any issue with having stubs, fakes, mocks, logs or screenshots inside test folder.

vpratfr:
Well, having the `routes` folder inside the `app` directory in Laravel has never been an issue either. That said, since `app` became a PSR4 compliant folder, they decided to keep things clean and organised.

Not saying that those folders MUST be moved. Just saying that there is a way to tell the browser to store them somewhere else, but the artisan command does not honour that choice.

At the very least, I would not mind the command not cleaning up my custom folders if it did not simply crash because the default folders are missing (in which case it should just skip the cleanup step).

staudenmeir:
I added an `is_dir()` check to `purgeScreenshots()` and `purgeConsoleLogs()`: #577

theomessin:
Any update on this?

driesvints:
@theomessin you're free to send in a pr for this.

derekmd:
@driesvints: [The recent pull request](https://github.com/laravel/dusk/pull/775) to allow configurable directories was declined so this should probably be closed.

> However, the DuskCommand complains that the folders in tests/Browser do not exist when trying to clean them.

in the root issue post was fixed by staudenmeir 2 years ago.

driesvints:
@derekmd thanks.

PR #775, opened 2020-05-14
[6.x] Ability to set artifacts directories

dmitryuk:
Resolve https://github.com/laravel/dusk/issues/393


taylorotwell:
I see no big reason to make this configurable.

PR #777, opened 2020-05-17
Look for the default app name from the environment

shanecp:
Most projects will have the project's name by default on the Homepage.

So we can look for the APP_NAME instead of "Laravel" on with the ExampleTest.

So it will reduce 1 step of going back to the ExampleTest and changing it.


driesvints:
Feel free to do this in your own app.

PR #784, opened 2020-06-11
[6.x] Support executing closures outside of the current selector scope

### LINKED ISSUE #783: A way to break selector scope
When writing components that sometimes render elements outside of a certain scope I find myself needing to break the selector scope that I'm in. Take the popular [Select](https://material-ui.com/components/selects/) component from [Material UI's react library](https://material-ui.com/) for example. Portions of the component aren't necessarily within the same css selection scope. The resulting html looks something like this(abbreviated for comprehension).

```html
<div class="mui-select">
   <label>My Label</label>
   <div>...input display</div>
</div>

... other portions of the page

<div class="mui-menu">
  <ul>
     <li class="select-option">1</li>
     <li class="select-option">2</li>
     <li class="select-option">3</li>
  </ul>
</div>

```

Now if I define a [component class in dusk](https://laravel.com/docs/7.x/dusk#components) for this, I walk myself into a corner as everything in the component would be scoped into the `.mui-select` class. Afaik there wouldn't be any way to select the `.mui-menu` from the component class. If there is please let me know and I'd be happy to do that.

Right now however, I'm getting away with this by temporarily removing the scope then adding it back which works fine, but feels clunky.

```php

class SelectInput extends BaseComponent
{

   // selector info, elements

    public function selectFromDropdown(Browser $browser, string $value)
    { 
        $prefix = $browser->resolver->prefix;
        $browser->resolver->prefix = "";
        $browser->within('.mui-menu', function ($browser) use ($value) {
           // do unscoped stuff
        });
        $browser->resolver->prefix = $prefix;
    }
}

```

If that seems like a good way to access it, then maybe we could add a method to allow a way to break the scope. So in `Laravel\Dusk\Browser` we could add something like:

```php
// not sure what the name would be for such a function
public function breakScope(Closure $callback) {
    $prefix = $browser->resolver->prefix;
    $browser->resolver->prefix = "";
    $callback($this)
    $browser->resolver->prefix = $prefix;
}
```

Infamoustrey:
There are scenarios, particularly when writing logic to encapsulate components that don't necessarily have a linear structure in the dom, that you might want to run certain logic outside of your current selector scope. This function simply executes a closure outside of the current scope.

Consider the following html generated from a js component from a popular react framework:
```html
<div class="mui-select">
   <label>My Label</label>
   <div>...input display</div>
</div>

... other portions of the page

<div class="mui-menu">
  <ul>
     <li class="select-option">1</li>
     <li class="select-option">2</li>
     <li class="select-option">3</li>
  </ul>
</div>
```

To interact with this component before you had to do something like this: 

```php
public function selectFromDropdown(Browser $browser, string $value)
{ 
    $prefix = $browser->resolver->prefix;
    $browser->resolver->prefix = "";
    $browser->within('.mui-menu', function ($browser) use ($value) {
           // do unscoped stuff
     });
    $browser->resolver->prefix = $prefix;
}
```

Now the code will look something like this:

```php
public function selectFromDropdown(Browser $browser, string $value)
{ 
    $browser->elsewhere('.mui-menu', function ($browser) use ($value) {
        // do unscoped stuff 
    }); 
}
```

Closes #783

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -481,6 +481,32 @@ class Browser
     }
 
     /**
+     * Execute a Closure outside of the current browser scope.
+     *
+     * @param  string  $selector
+     * @param  \Closure  $callback
+     * @return $this
+     */
+    public function elsewhere($selector, Closure $callback)
+    {
+        $browser = new static(
+            $this->driver, new ElementResolver($this->driver, 'body ' . $selector)
+        );
+
+        if ($this->page) {
+            $browser->onWithoutAssert($this->page);
+        }
+
+        if ($selector instanceof Component) {
+            $browser->onComponent($selector, $this->resolver);
+        }
+
+        call_user_func($callback, $browser);
+
+        return $this;
+    }
+
+    /**
      * Set the current component state.
      *
      * @param  \Laravel\Dusk\Component  $component
--- a/tests/BrowserTest.php
+++ b/tests/BrowserTest.php
@@ -135,6 +135,19 @@ class BrowserTest extends TestCase
         });
     }
 
+    public function test_elsewhere_method()
+    {
+        $driver = m::mock(stdClass::class);
+        $browser = new Browser($driver);
+
+        $browser->with('prefix', function ($browser) {
+            $browser->elsewhere('.my-class', function($browser) {
+                $this->assertInstanceof(Browser::class, $browser);
+                $this->assertEquals('body .my-class', $browser->resolver->prefix);
+            });
+        });
+    }
+
     public function test_page_macros()
     {
         $driver = m::mock(stdClass::class);


src/Browser.php
```diff
@@ -480,6 +480,32 @@ public function with($selector, Closure $callback)
...
     }
 
+    /**
+     * Execute a Closure outside of the current browser scope.
+     *
+     * @param  string  $selector
+     * @param  \Closure  $callback
+     * @return $this
+     */
+    public function elsewhere($selector, Closure $callback)
```
driesvints, line 490: Probably best `elseWhere`

Infamoustrey: Is it two words?

taylorotwell:
Can you make a PR to laravel/docs to document this?

Infamoustrey:
Will do!

src/Browser.php
driesvints: Ah I mistook this.

Issue #799, opened 2020-08-18
ElementResolver's format() replaces the part of css selector matching other elements shortcut key.


dsamburskyi:
- Dusk Version: 6.4.1
- Laravel Version: 7.22.4
- PHP Version: 7.4.6
- Database Driver & Version:

### Description:
resolver->format() replaces the part of selector matching the element shortcut. 
What that means is when there's a 'registered' element with the key (for example "title"), any occurrence of that key ("title") in other elements selector would be replaced with the key's value, 
like
```
'title'      => '[data-testid=title]',
'subitle' => '[data-testid=subtitle]',
```
The 'subtitle' element selector would resolve to `[data-testid=sub[data-testid=title]], thus triggering "Invalid selector" error.

### Steps To Reproduce:
Have registered element shortcuts where one element key is a part of another element selector.

driesvints:
I don't see an easy solution to this other than to make use of more specific selectors in your tests sorry. If you have a practical solution we'd appreciate a PR.

PR #802, opened 2020-08-25
[6.x] Fix resolver format() replacing a fragment of selector

dsamburskyi:
The PR addresses an issue #799 where resolver->format() replaces the part of selector matching the element shortcut.
What that means is when there's a 'registered' element with the key (for example "title"), any occurrence of that key ("title") in other elements selector would be replaced with the key's value, like
```
'title' => '[data-testid=title]',
'subtitle' => '[data-testid=subtitle]',
```
The 'subtitle' element selector would resolve to `[data-testid=sub[data-testid=title]]`, thus triggering "Invalid selector" error.

There's a logic added to check whether $selector is a string or not, as it can be an instance of the Component.
Updated/added tests for the relevant functionality.

### DIFF
--- a/src/ElementResolver.php
+++ b/src/ElementResolver.php
@@ -405,9 +405,15 @@ class ElementResolver
             return strlen($key);
         })->toArray();
 
-        $selector = str_replace(
-            array_keys($sortedElements), array_values($sortedElements), $originalSelector = $selector
-        );
+        $originalSelector = $selector;
+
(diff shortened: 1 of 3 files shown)


taylorotwell:
Maybe prefix your shorthand selectors with something? Like an @ sign or some other character.

dsamburskyi:
Do you consider the current behavior to be expected?

PR #815, opened 2020-09-17
[6.x] fix the `fitContent()` method

browner12:
well, my bad on this one....

In this PR of mine (https://github.com/laravel/dusk/pull/731) I screwed up the logic when flattening the conditional, and basically prevented all content fitting from occurring.

someone please double check my work on this one :sweat_smile:

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -284,7 +284,7 @@ class Browser
 
         $html = $this->driver->findElement(WebDriverBy::tagName('html'));
 
-        if (! empty($html) && ($html->getSize()->getWidth() <= 0 || $html->getSize()->getHeight() <= 0)) {
+        if (! empty($html) && $html->getSize()->getWidth() >= 0 && $html->getSize()->getHeight() >= 0) {
             $this->resize($html->getSize()->getWidth(), $html->getSize()->getHeight());
         }
 


SjorsO:
I noticed this method didn't work a while back, but I blamed my HTML, haha

I just tested the fix in this PR, I can confirm it solves the problem

driesvints:
@browner12 I think this re-introduced the bug originally fixed in https://github.com/laravel/dusk/pull/730 where the browser can be resized to zero width or height.

It probably should be:

```php
if (! empty($html) && $html->getSize()->getWidth() > 0 && $html->getSize()->getHeight() > 0) {
```

browner12:
That change seems to make sense.

Is the problem that `>=` and `<=` have no type strict options, and `getWidth()` or `getHeight()` are returning a falsy value?

driesvints:
No the problem is the `=` part that you kept. We explicitly want to prevent resizing when the width or height is zero or lower.

browner12:
Yah, that makes sense to me.

My only confusion is, if the height of the content is actually zero, what does it matter if we resize it to zero?  Is there a scenario where `getHeight()` is returning 0, but the height isn't actually zero?

driesvints:
Hmm, that I don't know myself I have to admit.

PR #816, opened 2020-09-22
[6.x] add Chrome driver auto detection

browner12:
Using the `--detect` flag, the command will see if it can determine the install driver version, and then install the latest of that major version.

All credit goes to @staudenmeir. This code was taken from https://github.com/staudenmeir/dusk-updater.

In our next major release, I feel like this should be the default option, and we should add a `--latest` flag to get back to the current default behavior.

### DIFF
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -304,7 +304,7 @@ class ChromeDriverCommand extends Command
 
             preg_match('/(\d+)(\.\d+){3}/', $process->getOutput(), $matches);
 
-            if (!isset($matches[1])) {
+            if (! isset($matches[1])) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


taylorotwell:
What does this do exactly? What do you mean by "determine the install driver version"?

browner12:
sorry, should have said "the installed chromium version".

this checks the currently installed version of Chromium, and uses that to install the appropriate Driver version.

for example, on linux, it would run `chromium-browser --version`.

So let's say the installed version is 84.0.123.45.  `php artisan dusk:chrome-driver --detect` would install the appropriate driver for v84.

driesvints:
@browner12 can you maybe also send in a PR to the docs for this?

browner12:
sure thing

PR #817, opened 2020-09-26
added waitForTextIn function

KaiReichart:
Added Method to wait for text in selector.

```php
public function waitForTextIn($selector, $text, $seconds = null)

```
Example usage:

```php
$this->browse(function(Browser $browser) use($user){
    $browser->loginAs($user)
        ->visit('/datatable')
        ->waitForTextIn('@datatableColStatus', STATUS_DONE, 60);
});
```

### DIFF
--- a/src/Concerns/WaitsForElements.php
+++ b/src/Concerns/WaitsForElements.php
@@ -111,6 +111,25 @@ trait WaitsForElements
     }
 
     /**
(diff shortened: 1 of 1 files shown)


driesvints:
This can be sent to the current stable branch. Also not sure why you sent it in as a draft pr?

PR #826, opened 2020-10-04
[6.x] Fix `logout()` when using AuthenticateSession middleware

### LINKED ISSUE #825: Logout issue when using AuthenticateSession middleware
- Dusk Version: v6.7.0
- Laravel Version: v8.8.0
- PHP Version: v7.3.3
- Database Driver & Version: //

### Description:
Can't login the user twice if the application is using AuthenticateSession middleware (```\Illuminate\Session\Middleware\AuthenticateSession```) after using the ```->logout()``` Dusk method.

### Steps To Reproduce:
- Create an application with a login page that uses the ```web``` middleware group.
- In the ```app/Http/Kernel.php``` file uncomment ```\Illuminate\Session\Middleware\AuthenticateSession::class```
- Write a test like this:
```php
// 1
    $browser
        ->visit(new LoginPage())
            ->type('@username', $this->credentials['email'])
            ->type('@password', $this->credentials['password'])
            ->click('@submit')
            ->assertPathIs('/admin/dashboard')
            ->assertAuthenticatedAs($this->admin, 'admin')
            ->logout('admin');
});

// 2
    $browser
        ->visit(new LoginPage())
            ->type('@username', $this->credentials['email'])
            ->type('@password', $this->credentials['password'])
            ->click('@submit')
            ->assertPathIs('/admin/dashboard')
            ->visit(new LogoutPage())
            ->assertGuest('admin');
});
```
- In the ```// 2``` test the request passes through https://github.com/laravel/framework/blob/8.x/src/Illuminate/Session/Middleware/AuthenticateSession.php#L54 and since the middleware sees that the ```'password_hash_' . $guard``` in the session doesn't match with the ```$request->user()->getAuthPassword()``` it tries to logout the user again before the new login happens.

EmanueleCoppola:
Should fix #825

### DIFF
--- a/src/Http/Controllers/UserController.php
+++ b/src/Http/Controllers/UserController.php
@@ -3,6 +3,7 @@
 namespace Laravel\Dusk\Http\Controllers;
 
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Session;
 use Illuminate\Support\Str;
 
 class UserController
@@ -56,6 +57,8 @@ class UserController
     public function logout($guard = null)
     {
         Auth::guard($guard ?: config('auth.defaults.guard'))->logout();
+
+        Session::forget('password_hash_'.$guard ?: config('auth.defaults.guard'));
     }
 
     /**


driesvints:
We don't support two the same browser calls in the same test method. Dusk is meant to mimic a real life scenario and you only do one single request when navigating to a page and running assertions.

taylorotwell:
@driesvints we don't?

driesvints:
@taylorotwell calling HTTP requests twice in the same test method has always lead to edge cases and problems afaik.

driesvints:
FYI: this wasn't related to double HTTP calls in the end.

Issue #832, opened 2020-11-02
[6.x] assertVue no longer works with Vue 3.x


mattkingshott:
- Dusk Version: 6.8.0
- Laravel Version: 8.12.3
- PHP Version: 7.4.11
- Database Driver & Version: N/A

### Description:

The `assertVue()`, `assertVueIsNot`, `assertVueContains` and `assertVueDoesNotContain` methods no longer work with Vue 3.x. This is due to the `__vue__` attribute no longer being present on components.

To address this issue, the `vueAttribute` method within the `MakesAssertions` trait will need to be rewritten to use an alternative means of extracting state from the component.

The following code will work with Vue 2.x and 3.x:

```php
public function vueAttribute($componentSelector, $key)
{
    $fullSelector = $this->resolver->format($componentSelector);

    return $this->driver->executeScript(
        "return document.querySelector('".$fullSelector."').__vue__?.".$key." || JSON.parse(JSON.stringify(document.querySelector('".$fullSelector."').__vueParentComponent.data)).".$key
    );
}
```

Note the use of JSON stringifying and parsing. This is because in Vue 3.x, the data object is a `Proxy`, so a deep clone is advisable to prevent any issues with nested properties / objects.

The only reason that I'm not submitting this as a PR is because I'm still fairly new to Vue 3.x and there may be situations where accessing the state in the above manner is not advisable. Otherwise, it should be good to merge in.

driesvints:
Also not familiar with Vue so need some help with this one.

mattkingshott:
Probably advisable given that my setup uses `AsyncComponentWrappers` via `defineAsyncComponent()`, which might explain why I have to access data through `__vueParentComponent` instead of `__vnode`.

That said, I'm pretty much guessing at this point...

derekmd:
I just submitted a pull request that will fix the issue. A couple things I found trying out the above expression:

1. `el.__vueParentComponent.ctx` should be used over `el.__vueParentComponent.data` for a wider scope of component values that can be bound to `v-model` `<input>` elements, output `<div>`s, etc.
2. The given ternary expression would break Vue 2 assertions on falsy values like JavaScript `false` or string value `'0'`:
   ```javascript
   return '0' || JSON.parse(JSON.stringify(document.querySelector('".$fullSelector."').__vueParentComponent.data)).".$key
   ```
   If the Vue prop is falsy, the above expression would fallback to the Vue 3 expression that would cause an exception when Vue 2 is installed.

[...]

PR #834, opened 2020-11-19
[6.x] Add Vue 3 support to the assertVue*() methods

derekmd:
Fixes: https://github.com/laravel/dusk/issues/832

Vue 3 no longer adds a `__vue__` property to DOM nodes to allow querying for props. Instead the `__vueParentComponent` must be inspected. Its 'ctx' property is an object containing the:

* passed in component props
* `data()` defined within the component
* any computed props

The above three items would be checked by Laravel Dusk to assert `v-model` value bindings.

When Dusk is looking up the Vue prop value of a DOM node, it will first look for a Vue 2 `__vue__` property, then fallback to `__vueParentComponent.ctx` for Vue 3.

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -876,7 +876,10 @@ JS;
         $fullSelector = $this->resolver->format($componentSelector);
 
         return $this->driver->executeScript(
-            "return document.querySelector('".$fullSelector."').__vue__.".$key
+            "var el = document.querySelector('".$fullSelector."');".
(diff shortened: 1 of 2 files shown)


taylorotwell:
This would have been hard to figure out on my own. Thank you ❤️

driesvints:
Thanks @derekmd!

lk77:
@derekmd For me, this vue3 support does not always work, it can cause `TypeError: Converting circular structure to JSON`
because of the JSON.stringify

PR #843, opened 2020-12-13
[6.x] Add new assertions `assertSeeEmptyTextIn()` and `assertDontSeeEmptyTextIn()`

u01jmg3:
Sometimes it is helpful to check that a selector has some text (or conversely has no text) particularly when the value of the text is dynamically populated. This means it's less about asserting what the value of the text is but rather that there is some text (i.e. the contents of the selector is not empty).

This is where `assertSeeEmptyTextIn()` and `assertDontSeeEmptyTextIn()` comes into play.

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -206,6 +206,46 @@ trait MakesAssertions
     }
 
     /**
+     * Assert that the given selector has no text.
+     *
+     * @param  string  $selector
+     * @return $this
+     */
+    public function assertSeeEmptyTextIn($selector)
+    {
+        $fullSelector = $this->resolver->format($selector);
+
+        $element = $this->resolver->findOrFail($selector);
+
+        PHPUnit::assertTrue(
+            $element->getText() === '',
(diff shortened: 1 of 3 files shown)


taylorotwell:
Can't you pass an empty string to assertSeeIn and assertDontSeeIn?

u01jmg3:
@taylorotwell: nope, `assertSeeIn()` and `assertDontSeeIn()` both use `\Str::contains()`.

```php
dump(Str::contains('text', '')); // false

dump(Str::contains('', '')); // false
```

u01jmg3:
@driesvints: any chance this can be looked at again? Apologies but I don't know what the procedure is for getting Taylor to take another look at a closed PR once he's made a comment.

I've given an explanation above as to why I believe `assertSeeIn()` and `assertDontSeeIn()` can't be used for checking if a selector has empty text or not. Unless I'm missing something...

driesvints:
Re-opened this for Taylor to review.

taylorotwell:
Renamed to `assertSeeAnythingIn` and `assertSeeNothingIn`.

PR #847, opened 2020-12-20
Use $browser given from closure to received scoped selector.

crynobone:
Additional fix to #846 PR.


driesvints:
Thanks!

PR #852, opened 2021-01-18
[6.x] Add setInputValue alias

SjorsO:
This PR adds a `setInputValue` alias that makes it easier to discover this functionality. Setting input values directly is especially useful when working with date inputs in React applications. Typing the date in an input can cause weird problems.

The method is currently very difficult to discover inside your IDE. Any search for "input" or "set" won't yield any results. Only by thoroughly reading the documentation you can find out you're supposed to use the `value` method. (I only discovered it while I was making the `setInputValue` method for this PR).

Before:
image

After:
image

### DIFF
--- a/src/Concerns/InteractsWithElements.php
+++ b/src/Concerns/InteractsWithElements.php
@@ -73,6 +73,18 @@ trait InteractsWithElements
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hmm, I dunno - it's right there in the docs? 

image

Issue #856, opened 2021-01-23
fitToContent() breaks screenshots if document height/width is 0


HugoJF:
- Dusk Version: 6.11.1
- Laravel Version: 8.24.0
- PHP Version: 7.4.14
- Database Driver & Version: MySQL 8.0.22
- Chrome version: 88.0.4324.96

### Description:
TL;DR: It's basically the same issue described in #729 (that issue was the only reason I figured what was happening).

I have a somewhat weird layout for one application I'm playing around and I couldn't get `assertDontSee` to fail, they always hanged the whole test if the assertion failed. After checking that issue, I ran the chrome-driver in verbose mode, and saw Dusk was resizing my windows to `height: 0` and after that, every screenshot would timeout the entire test.

```
[1611405228.586][INFO]: [39604110c03f03139b9e8d47ee279816] RESPONSE SetWindowRect {
   "height": 0,
   "width": 1280,
   "x": 0,
   "y": 0
}
[1611405228.588][INFO]: [39604110c03f03139b9e8d47ee279816] COMMAND Screenshot {

}
[1611405228.591][DEBUG]: DevTools WebSocket Command: Page.captureScreenshot (id=232) A03ED0ACB90A15BC68710AAE6C147D9D {

}
[1611405238.613][SEVERE]: Timed out receiving message from renderer: 10.000
[1611405238.618][WARNING]: screenshot failed, retrying timeout: Timed out receiving message from renderer: 10.000
```

Adding `min-height: 100vh` to by body fixed this issue (or anything that would report `height > 0`).

### Steps To Reproduce:
Same as #729 

1. Visit any page that has it's height reported as zero
2. Intentionally fail the test within the code
3. Dusk CLI hangs for several minutes and finally aborts with an timeout exception (may vary) because the driver hangs

---

The whole problem is somewhat simple to fix (I'm not sure what #730 is fixing given I'm having the same problems):

1. Don't resize if any dimension is 0;
2. Don't resize below a given resolution (maybe the original window size);
3. Clamp resize resolution (original window size or 1);
4. Add a warning when any dimensions is 0.

This problem was pretty annoying to debug (2 days messing around with Homestead), at the very least sugestion 3 or 4 would point other devs to the correct direction (fix their layouts reporting 0 height) without making any significant changes.

driesvints:
Seems like a different PR re-introduced this. I've sent in a fix for it here: https://github.com/laravel/dusk/pull/858

RaZeR-RBI:
[...]

PR #858, opened 2021-01-25
[6.x] Do not resize to zero

driesvints:
At the moment fitContent will be able to resize to zero and thus making assertions basically useless. There was a PR that fixed this https://github.com/laravel/dusk/pull/730 but https://github.com/laravel/dusk/pull/815 seemed to have re-introduced it.

Fixes https://github.com/laravel/dusk/issues/856


browner12:
thanks for stepping in on this!

PR #861, opened 2021-01-29
[7.x] Drop PHPUnit v8

driesvints:
This PR drops support for PHPUnit v8 which is unmaintained as of next week.


taylorotwell:
@driesvints has some conflicts I'll let you resolve.

PR #864, opened 2021-02-06
[7.x] Add ability to toggle headless options using --browse command options.

crynobone:
This follows GitHub Action environment where `CI=true` is always available.

* By default, `php artisan dusk` will run as headless by appending CI=true environment.
* This is equivalent to `CI=true vendor/bin/phpunit -c phpunit.xml` (users not using artisan dusk will see this as a breaking change).
* By adding `--browse`, the command will run by opening a browser

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -17,7 +17,8 @@ class DuskCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'dusk {--without-tty : Disable output to TTY}';
+    protected $signature = 'dusk {--without-tty : Disable output to TTY}
+                                {--browse : Display browser output instead of using headless mode}';
 
     /**
      * The console command description.
@@ -58,12 +59,21 @@ class DuskCommand extends Command
 
         $this->purgeSourceLogs();
 
-        $options = array_slice($_SERVER['argv'], $this->option('without-tty') ? 3 : 2);
+        $options = array_slice(
+            $_SERVER['argv'],
+            ($this->option('without-tty') && $this->option('browse') ? 4 : $this->option('without-tty') || $this->option('browse')) ? 3 : 2
+        );
 
         return $this->withDuskEnvironment(function () use ($options) {
+            $env = [];
+
+            if (! $this->option('browse')) {
+                $env['CI'] = true;
+            }
+
             $process = (new Process(array_merge(
                 $this->binary(), $this->phpunitArguments($options)
-            )))->setTimeout(null);
+            ), null, array_filter($env)))->setTimeout(null);
 
             try {
                 $process->setTty(! $this->option('without-tty'));
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -59,10 +59,10 @@ class DuskCommand extends Command
 
         $this->purgeSourceLogs();
 
-        $options = array_slice(
-            $_SERVER['argv'],
-            ($this->option('without-tty') && $this->option('browse') ? 4 : $this->option('without-tty') || $this->option('browse')) ? 3 : 2
-        );
+        $options = collect(array_slice($_SERVER['argv'], 2))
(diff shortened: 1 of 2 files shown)


src/Console/DuskCommand.php
```diff
@@ -58,12 +59,21 @@ public function handle()
...
+            ->filter(function ($option) {
+                return ! in_array($option, ['--without-tty', '--browse']);
+            })->values()->all();
 
         return $this->withDuskEnvironment(function () use ($options) {
+            $env = [];
+
+            if (! $this->option('browse')) {
+                $env['CI'] = true;
+            }
```
derekmd, line 72: Dusk uses headless by default so it may be more clear to name the environment variable with its `--browse` edge case description? Desktop test runs will also use headless mode (not in CI) until this new argument is used to analyze browser behaviour that the test failure screenshot doesn't capture.

Replacing this with the below code would help with packages keeping Laravel Dusk backwards capability:

```php
if ($this->option('browse')) {
    $env['DUSK_BROWSE'] = true;
}
```

```php
$options = (new ChromeOptions)->addArguments(collect([
    '--window-size=1920,1080',
])->unless(isset($_SERVER['DUSK_BROWSE']) || isset($_ENV['DUSK_BROWSE']), function ($items) {
    return $items->merge([
        '--disable-gpu',
        '--headless',
    ]);
})->all());
```

e.g., this `--browse` feature would be introduced in laravel/dusk 7.x so custom stubs must be updated to support it.

From my laravel-dusk-firefox package:

https://github.com/derekmd/laravel-dusk-firefox/blob/58d10303c580b120d0f70cb4a202d188c89a192b/stubs/DuskTestCase.stub#L56-L82

If `--headless` is only used when `$env['CI']` exists, there isn't a way to make the custom stub have backwards compatibility with laravel/dusk 6.x releases that don't set `$env['CI']`. There would need to be separate stubs for each major Dusk version.

taylorotwell: @crynobone any thoughts on comment above?

taylorotwell:
@crynobone can you describe more the goal of this PR?

src/Console/DuskCommand.php
crynobone: Seem fine, just not sure if `DUSK_BROWSE` would be suitable. However both way still require stub TestCase to be updated in order to use `--browse`.

crynobone:
The goals are mainly:
* Ensure CI can run headless chrome without changing DuskTestCase.php
* Allows developer to run tests without headless and combine it with `$browser->tinker();` to help them when building tests.

src/Console/DuskCommand.php
derekmd: Maybe `DUSK_HEADLESS_DISABLED` is more clear? Whatever the name, the main goal would be for `DuskTestCase` to look for the `--browse` option be enabled rather than the absence of that option.

taylorotwell:
@derekmd are you able to PR your suggested approach so that I can see it?

crynobone:
The main reason this PR suggests `$_ENV['CI']` is that it's been a commonly accepted default environment variables for most CI:

* https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables
* https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
* https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

derekmd:
I guess I'm confused why `DuskCommand` is assigning `CI=true` when locally running arg-less `php artisan dusk` (the typical use) will be setting that variable when it not a continuous integration environment on a dev's machine. Should assigning `CI=true` be left to the environment configuration outside of Dusk?

I understand the thought behind `DuskTestCase` looking for `CI=true`, however out-of-the-box _Laravel Dusk is already CI-ready_. Given Laravel Dusk already runs in headless mode by default, does the package need to know whether a CI environment is currently active? It assumes that and only turns off headless mode when `--browse` is given.

PR #867, opened 2021-02-08
[7.x] Add Dusk command argument --browse

derekmd:
Alternate implementation of https://github.com/laravel/dusk/pull/864 as requested by Taylor.

Chromedriver headless mode can be disabled by calling:

```
php artisan dusk --browse
```

This allows devs to debug on desktop, showing the browser window as tests run. After upgrading to laravel/dusk 7.x, apps will need to reinstall the updated stub to use the `--browse` argument.

Differences in this implementation:

* shows in the code that headless mode is the default behavior, both in CI and _during local dev_
  * the other pull request sets `$env['CI'] = true` when Dusk is being run on laptop/desktop machines
* allows custom `DuskTestCase` stub classes to look for the presence of `--browse` rather than detecting its absence when choosing headless mode. Since Laravel Dusk <= 6.x stubs don't have this argument, it causes logic issues for 3rd-party packages that support many versions of Dusk.

Environment variable `DUSK_HEADLESS_DISABLED` could shortened but that's the best descriptive name I can currently think of.

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -17,7 +17,9 @@ class DuskCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'dusk {--without-tty : Disable output to TTY}';
+    protected $signature = 'dusk
+                {--browse : Display browser output instead of using headless mode}
+                {--without-tty : Disable output to TTY}';
 
     /**
      * The console command description.
@@ -58,12 +60,16 @@ class DuskCommand extends Command
 
         $this->purgeSourceLogs();
 
-        $options = array_slice($_SERVER['argv'], $this->option('without-tty') ? 3 : 2);
+        $options = collect($_SERVER['argv'])
+            ->slice(2)
+            ->diff(['--browse', '--without-tty'])
+            ->values()
+            ->all();
 
         return $this->withDuskEnvironment(function () use ($options) {
             $process = (new Process(array_merge(
                 $this->binary(), $this->phpunitArguments($options)
-            )))->setTimeout(null);
+            ), null, $this->env()))->setTimeout(null);
 
             try {
                 $process->setTty(! $this->option('without-tty'));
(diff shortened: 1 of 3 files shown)


taylorotwell:
@crynobone do you have any thoughts on this implementation? Would this solve your use case?

crynobone:
@taylorotwell It does solve "ability to toggle between headless mode on local machine"

However, the reason why I propose to use `CI` environment is to avoid bad configuration on CI environment. Using this PR approach, if CI environment uses `DUSK_HEADLESS_DISABLED` or run `php artisan dusk --browse` on CI environment they will accidentally run chrome without `--headless` and will see the following error:

```
Facebook\WebDriver\Exception\UnknownErrorException: unknown error: Chrome failed to start: exited abnormally.
  (unknown error: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
  ```
  
 The error isn't explicit and the developer will have to debug the issue manually. My PR avoid this because common CI predefine `CI` environment variable so `php artisan --browse` has no affect on CI environments:
 
 * https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables
* https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
* https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
* https://devcenter.heroku.com/articles/heroku-ci#immutable-environment-variables
* https://docs.gitlab.com/ee/ci/variables/predefined_variables.html
* ChipperCI supports it https://twitter.com/ChipperCI/status/1358967793723572226?s=20 but documentation not updated yet

**Comment by derekmd** [comment]
I've added a `CI=true` check for when `--browse` is mistakenly added to a CI .yml configuration:

```php
if ($this->option('browse') && ! isset($_ENV('CI')) && ! isset($_SERVER['CI'])) {
    return ['DUSK_HEADLESS_DISABLED' => true];
}
```

So this pull request now behaves the same as your pull request, ignoring `--browse` when `CI=true` exists (`DuskCommand` warning feedback could be added but that seems overkill.) However this branch gives custom Dusk stub classes more control to choose headless mode or not.

crynobone:
@derekmd 👍🏼 much better. 

@taylorotwell I think we should go with this PR.

taylorotwell:
Does this need to go to master or can it go to current branch @derekmd ?

derekmd:
Sure, it can target the 6.x branch since no existing method signatures are being changed. I assume Laravel doesn't have an extreme interpretation of semver where we're concerned about userland extending `DuskCommand` and have a custom `env()` method that conflicts with the new `protected` method I added.

6.x branch: https://github.com/laravel/dusk/pull/870

PR #871, opened 2021-02-12
[6.x] Fix DuskCommand syntax mistake

derekmd:
Now that I've actually tested the `--browse` feature in a sample app, fix a stupid syntax error I introduced in: https://github.com/laravel/dusk/pull/870

It would be good to get test coverage on `DuskCommand` but it may require:
* disabling `$this->withDuskEnvironment()`
* using Mockery static namespace overloading (eh) because the `Symfony\Component\Process\Process`  object isn't resolved from Laravel's container.

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -129,7 +129,7 @@ class DuskCommand extends Command
      */
     protected function env()
     {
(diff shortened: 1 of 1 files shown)


driesvints (approved):

driesvints:
Thanks

PR #873, opened 2021-02-12
[7.x] Remove chrome binaries

driesvints:
Remove shipped chrome binaries. This is because of a couple of reasons:

- They take up quite some space
- They often conflict with the installed Chrome version
- We now have a dedicated command to easily install the correct one with the correct version

The current `dusk:install` command already installs the correct Chrome version. I've also modified this to only install the OS specific one by default. Installing binaries which can't be run anyway doesn't really makes sense. The option remains on the `dusk:chrome-driver` command if people really need it.

This PR also fixes quite a few tests. I noticed at some point I was getting weird Mockery exception failures after introducing Testbench. Turns out none of the existing tests had the `m::close` in place. I added these and fixed all existing tests with the proper expectations.


driesvints:
@crynobone I think I need to reverse PHPUnit v10 support on Dusk since Testbench doesn't yet supports it?

PR #876, opened 2021-02-16
[6.x] Add support for Mac ARM64 architecture

crynobone:


src/OperatingSystem.php
```diff
@@ -35,4 +41,21 @@ public static function onMac()
...
+    {
+        switch (php_uname('m')) {
+            case 'arm64':
+                return 'mac-arm';
+            case 'x86_64':
+                return 'mac-intel';
+            default:
+                return 'mac';
+        }
+    }
```
crynobone, line 60: This different from PR sent to `master`. Instead of having `onMacIntel()` and `onMacArm()` we can just stick with `onMac()` and then have a new method to return architecture value.

src/Console/ChromeDriverCommand.php
```diff
@@ -59,6 +59,8 @@ class ChromeDriverCommand extends Command
     protected $slugs = [
         'linux' => 'linux64',
         'mac' => 'mac64',
+        'mac-intel' => 'mac64',
+        'mac-arm' => 'mac64_m1',
```
crynobone, line 63: `mac` key will remain for fallback in case architecture value doesn't match `x86_64` or  `arm64`.

driesvints:
@crynobone this looks good to me 👍 

@sirdharma can you give this a review? We're planning on introducing ARM64 support on Dusk 6.x so we can delay Dusk 7.x until PHPUnit 10 is released.

sirdharma:
@driesvints LGTM 👍

PR #879, opened 2021-02-23
[6.x] Add `assertNotPresent()` assertion

GingerNinjaNicko:
I recently ran into a use case where I needed to ensure an array of inputs would definitely submit on the page, and another array of inputs definitely were not.

`assertPresent()` was exactly what I needed, but I could not find an inverse of it.
`assertMissing()` checked that the input was visible, but some of the inputs are hidden and replaced via JavaScript, so this doesn't work.

Therefore, I added an `assertNotPresent()` assertion for this specific use-case. 
I did think of naming it `assertAbsent()` but think the distinction between this and `assertMissing()` could become confusing...

This is the relevant part of the test I wanted to write:
```php
collect($this->data['support_staff_only_fields'])->each(
    fn($field) => $browser->assertPresent("[name='{$field}']")
);
collect($this->data['franchisee_only_fields'])->each(
    fn($field) => $browser->assertNotPresent("[name='{$field}']")
);
```

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -719,6 +719,24 @@ JS;
 
         return $this;
     }
+    
+    /**
+     * Assert that the element matching the given selector is not present.
+     *
+     * @param  string  $selector
+     * @return $this
+     */
+    public function assertNotPresent($selector)
+    {
+        $fullSelector = $this->resolver->format($selector);
+
+        PHPUnit::assertTrue(
+            is_null($this->resolver->find($selector)),
+            "Element [{$fullSelector}] is present."
+        );
+
+        return $this;
+    }
 
     /**
      * Assert that the element matching the given selector is not visible.
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -719,7 +719,7 @@ JS;
 
         return $this;
     }
-    
+
     /**
      * Assert that the element matching the given selector is not present.
      *
--- a/tests/MakesAssertionsTest.php
+++ b/tests/MakesAssertionsTest.php
@@ -669,6 +669,37 @@ class MakesAssertionsTest extends TestCase
         }
     }
 
+    public function test_assert_not_present()
+    {
+        $driver = m::mock(stdClass::class);
+        $element = m::mock(stdClass::class);
+
+        $resolver = m::mock(stdClass::class);
+        $resolver->shouldReceive('format')->with('foo')->andReturn('body foo');
+        $resolver->shouldReceive('find')->with('foo')->andReturn(
+            null,
+            null
+        );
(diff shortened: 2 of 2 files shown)


taylorotwell:
So under what circumstances is something "not present"? When it is in the source but not visible? Some other scenario?

GingerNinjaNicko:
No, something is "not present" when it cannot be found in the source and therefore _won't_ be visible.

Currently, there are the following assertions which descriptions from the docs:
- `assertVisible()`: Assert that the element matching the given selector _is visible_
- `assertPresent()`: Assert that the element matching the given selector _is present_

To ensure something _is **not** visible_ I can use:
- `assertMissing()`: Assert that the element matching the given selector _is **not** visible_

But there is no method I can see that allows me to ensure an element _is **not** present_

In my case, I am checking the presence of inputs, some of which may be hidden & replaced by javascript for a WYSIWYG editor. 
I need to ensure that some of these inputs do not appear in the source, so they don't post along with the form.

taylorotwell:
`assertMissing` checks that something is either not visible OR missing entirely from the source. Will that not be sufficient for your use case?

GingerNinjaNicko:
That was the behaviour I was expecting, but unfortunately, that's not the case. 
It only checks that the element is not displayed.

This is the `assertMissing` function code:
```php
public function assertMissing($selector)
{
    $fullSelector = $this->resolver->format($selector);

    try {
        $missing = ! $this->resolver->findOrFail($selector)->isDisplayed();
    } catch (NoSuchElementException $e) {
        $missing = true;
    }

    PHPUnit::assertTrue(
        $missing,
        "Saw unexpected element [{$fullSelector}]."
    );

    return $this;
}
```

GingerNinjaNicko:
Rather than create a breaking change by modifying `assertMissing` I think an extra function would make more sense.

taylorotwell:
@GingerNinjaNicko but in the very code you posted, it looks like it catches the exception for a totally missing element and also sets that to $missing = true?

GingerNinjaNicko:
Ah yes apologies @taylorotwell, I think you were right with your first comment but I got a little confused...

The difference is what causes the assertion to fail.
- `assertMissing`: passes when the element is not visible, even if it exists within the source. 
- `assertNotPresent`: fails when the element is not present in the source and so also not visible.

Hopefully I'm making more sense now.

SjorsO:
Maybe `assertSourceMissingSelector` is a better name for this new method? There is already an `assertSourceMissing` assertion, but that accepts a string instead of a selector.

taylorotwell:
@SjorsO yeah that might be a good idea... @GingerNinjaNicko could you rename the method to that and put it by the other assertSourceMissing method?

taylorotwell:
Eh `assertNotPresent` is fine I guess.

PR #887, opened 2021-03-22
[6.x] Add `purge` console command

crynobone:
image

#### Purging when folder doesn't exist.

image

### DIFF
--- a/src/Console/PurgeCommand.php
+++ b/src/Console/PurgeCommand.php
@@ -1,0 +1,121 @@
+<?php
+
+namespace Laravel\Dusk\Console;
+
+use Dotenv\Dotenv;
+use Illuminate\Console\Command;
+use Illuminate\Support\Str;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Process\Exception\ProcessSignaledException;
+use Symfony\Component\Process\Exception\RuntimeException;
+use Symfony\Component\Process\Process;
+
+class PurgeCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'dusk:purge';
+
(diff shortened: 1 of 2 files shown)


driesvints (approved):

taylorotwell:
So we don't want to purge everytime you run `dusk` anymore?

crynobone:
@taylorotwell on `dusk`, it will silently call `dusk:purge`

driesvints:
I realized that technically there are breaking changes here since the protected methods in the Dusk command are now gone. We'll need to be aware of that.

crynobone:
@driesvints ahhh yeah, we could revert `DuskCommand` changes and only add the new command for 6.x

taylorotwell:
@crynobone sounds good

crynobone:
@driesvints @taylorotwell done.

src/Console/PurgeCommand.php
```diff
@@ -0,0 +1,116 @@
...
+     * @param  string  $patterns
+     * @return void
+     */
+    protected function purgeDebuggingFiles($relativePath, $patterns)
+    {
+        $path = base_path($relativePath);
+
+        if (! is_dir($path)) {
+            $this->warn(
+                "Ignored purging none existing [{$relativePath}] path.", OutputInterface::VERBOSITY_DEBUG
```
derekmd, line 100: Since English likes to play games with us, this is a homonym of the more grammatically-correct "non-existent".

The framework's `Filesystem::get()` method has a similar exception message so maybe the two can be combined:

https://github.com/laravel/framework/blob/b3bea36b9e3a5b6e5b1d12e2ddc259beff656982/src/Illuminate/Filesystem/Filesystem.php#L51-L58

> Purge skipped the directory that does not exist at [{$relativePath}] path.

crynobone: Updated, thanks @derekmd

PR #888, opened 2021-04-02
[6.x] Added middleware configuration for authentication endpoints

thijskok:
In some cases, the 'web' middleware doesn't quite cut it when you're trying to authenticate against Dusk, for example:

- You have set up Laravel Fortify with custom middleware.
- You're using a multi-tenancy approach where the middleware needs to identify the tenant before authentication can take place.

The suggested configuration is optional and shouldn't break anything.

Current workaround would be to override the service provider and rewrite the routing code yourself (there are quite few lines to override though).


thijskok:
@taylorotwell Thanks!

PR #891, opened 2021-04-05
[7.x] Update `scrollTo` to work without JQuery

amir9480:


taylorotwell:
I don't see a big reason to do this tbh.

amir9480:
@taylorotwell 
For projects that using JQuery, it's not a big reason.
But what about projects that don't use JQuery

PR #913, opened 2021-07-28
Make `DuskCommand` compatible with Pest

freekmurze:
This PR introduces a flag will allow  Pest to be used, so Dusk tests can be written with the Pest syntax.

### DIFF
--- a/src/Console/DuskCommand.php
+++ b/src/Console/DuskCommand.php
@@ -20,7 +20,7 @@ class DuskCommand extends Command
     protected $signature = 'dusk
                 {--browse : Open a browser instead of using headless mode}
                 {--without-tty : Disable output to TTY}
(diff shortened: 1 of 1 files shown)


freekmurze:
I've tested out the changes in a real app, and @nunomaduro also signed off on this one 👍

driesvints:
Instead of a `--pest` flag, would a `--binary` flag here be of more use? That way test runners other than Pest and PHPUnit can also be used (if there are any). You could still alias the binary flag in a local alias.

freekmurze:
I talked to Nuno about this as well. We decided on a `--pest` flag zo users don't need to know the location of the Pest binary. There aren't many other test runners out there that are asking for Dusk support, so we wanted to make it as easy for Pest users as possible.

Should other test runners emerge, a binary flag could still be added.

PR #919, opened 2021-08-30
[6.x] Set Component as possible variable type to phpdoc of methods with, within, elsewhere

dmitryuk:
phpstan throws an error on lines above.
PR makes possible to set `Component` class as a variable to methods `with`, `within`, `elsewhere` without hints.


driesvints:
Please use FQN in all DocBlocks.

dmitryuk:
> Please use FQN in all DocBlocks.

Done

Issue #920, opened 2021-09-01
dusk:fails error with message The "pest" option does not exist.


SupianIDz:
- Dusk Version: 6.17.1
- Laravel Version: 8.56.0
- PHP Version: 8.0.10
- Database Driver & Version: MySQL 8.0.26

### Description:
According to the documentation [running-test](https://laravel.com/docs/8.x/dusk#running-tests), to save testing time, I can use the `php artisan dusk:fails` command to rerun the test.

But when I run the command, I get the message ` The "pest" option does not exist. ` even though the `pestphp/pest-plugin-laravel` plugin is properly installed, and works fine when I run the command `php artisan dusk --pest`

### Steps To Reproduce:

1. `composer create-project larave/laravel test`
2. `cd test`
3. `configure .env`
4. `composer require --dev laravel/dusk`
5. `php artisan dusk:install`
5. `composer require --dev pestphp/pest-plugin-laravel`
6. `php artisan pest:install`
7. `php artisan dusk --pest`

```php
$this->browse(function (Browser $browser) {
    $browser
       ->visit('/')
       ->assertSee('Laravel');
});
```
![1](https://user-images.githubusercontent.com/37969970/131601953-ade69950-2cc7-426d-9eda-bc5e1aec6d12.png)

and make the test an error 

![2](https://user-images.githubusercontent.com/37969970/131602168-36f96090-dd9d-4778-b041-d759567a014e.png)

Then fix the test and run `php artisan dusk:fails`

and I get an error like this:

![3](https://user-images.githubusercontent.com/37969970/131602313-bdf26b65-db1f-456f-98aa-e8b22bbb729c.png)

driesvints:
Heya, thanks for reporting.

We'll need more info and/or code to debug this further. Can you please create a repository with the command below, commit the code that reproduces the issue as separate commits on the main/master branch and share the repository here? Please make sure that you have the [latest version of the Laravel installer](https://github.com/laravel/installer) in order to run this command. Please also make sure you have both Git & [the GitHub CLI tool](https://cli.github.com) properly set up.

```bash
laravel new bug-report --github="--public"
```

Please do not amend and create a separate commit with your custom changes. After you've posted the repository, we'll try to reproduce the issue.

Thanks!

PR #921, opened 2021-09-03
[6.x] Fix "pest" option does not exist on dusk:fails command

SupianIDz:
This is to fix the [#920] issue.

This is because the $signature property of the parent class is overridden by the child class for which the "pest" option is missing.

See :
https://github.com/laravel/dusk/blob/6.x/src/Console/DuskCommand.php#L102
https://github.com/laravel/dusk/blob/6.x/src/Console/DuskCommand.php#L20
https://github.com/laravel/dusk/blob/6.x/src/Console/DuskFailsCommand.php#L12

### DIFF
--- a/src/Console/DuskFailsCommand.php
+++ b/src/Console/DuskFailsCommand.php
@@ -9,9 +9,10 @@ class DuskFailsCommand extends DuskCommand
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


driesvints (approved):

SupianIDz:
Thanks

PR #924, opened 2021-09-14
Fix DevToolsActivePort file doesn't exist when Dusk run as "root"

SupianIDz:
This happens when Dusk is running in an environment with the user "root" because Chrome doesn't allow running as "root" without `--no-sandbox`.

See : https://crbug.com/638180

error

### DIFF
--- a/stubs/DuskTestCase.stub
+++ b/stubs/DuskTestCase.stub
@@ -33,6 +33,7 @@ abstract class DuskTestCase extends BaseTestCase
     {
         $options = (new ChromeOptions)->addArguments(collect([
             '--window-size=1920,1080',
+            '--no-sandbox',
         ])->unless($this->hasHeadlessDisabled(), function ($items) {
             return $items->merge([
                 '--disable-gpu',


taylorotwell:
Add to your own stub if you need it.

joaopalopes24:
@SupianIDz How did you solve this problem?

joaopalopes24:
Even adding `--no-sandbox`, I still have the error.

SupianIDz:
> @SupianIDz How did you solve this problem?

can you attach the error log?

joaopalopes24:
image

I just managed to run it, but I have to leave this command running in my terminal:

`./vendor/laravel/dusk/bin/chromedriver-linux --log-level=ALL`

SupianIDz:
@joaopalopes24 

can you run `./opt/google/chrome/chrome --no-sandbox` and `./opt/google/chrome/chrome`  via terminal ?

joaopalopes24:
Yes, it's works:

`/opt/google/chrome/chrome --no-sandbox`

SupianIDz:
For a long time, I only needed the `--no-sandbox` flag only when the active session was the `root` user on Linux and had no problem on regular users. 

Unfortunately I can't help much, because I don't use Linux anymore. You can try using a regular user (non-root user) as the final suggestion, if you still have problems, try asking Stackoverflow or other forums.

joaopalopes24:
I appreciate the help. I managed to partially resolve it and will continue testing other fixes here. Thanks.

SupianIDz:
> I appreciate the help. I managed to partially resolve it and will continue testing other fixes here. Thanks.

@joaopalopes24 or you can try with `--headless` flag.

joaopalopes24:
I tried, but the error persists. The only way I could get it to run was to leave the command I sent running in the terminal next to the tests.

PR #926, opened 2021-09-21
Chnage assertAttribute to use assertStringContainsString

samuelhgf:
This PR make change in assertAttribute method to assertStringContainsString instead of assertEquals.

This will provide us to assert if a class exists inside a class attribute that most of the time has more than one class separated by spaces, for example.


taylorotwell:
Add a new `assertAttributeContains` method please. Can you send a separate PR for that?

Issue #633, opened 2019-03-27
Incomplete list of element types for function "InputValue"


ostapski:
- Dusk Version: 5.02
- Laravel Version: 8
- PHP Version: 7.1.27
- Database Driver & Version:

### Description:
Function "InputValue missing "SELECT" type of the element.
When used on a "SELECT" - function->getText() is called instead of ->getAttribute()
returning all options of select (instead of 1)

![201903276cld7no3](https://user-images.githubusercontent.com/2835357/55062345-1c241300-5076-11e9-90fd-195a7c9d8109.png)

as well .... - Text area - should not be in the list ....

```
    public function inputValue($field)
    {
        $element = $this->resolver->resolveForTyping($field);
        return in_array($element->getTagName(), ['input', 'textarea'])
                        ? $element->getAttribute('value')
                        : $element->getText();
    }
```

### Steps To Reproduce:

Call ->assertInputValue($field, $value) 
On "SELECT" type of element (with more than 1 options inside)

staudenmeir:
> Text area - should not be in the list

See #237.

driesvints:
Isn't this more of a case where the wrong assertion is used? Can you please post your entire test?

ostapski:
> 
> 
> Isn't this more of a case where the wrong assertion is used? Can you please post your entire test?

is "assertInputValue" not meant to be used on selects ?
calling "getText()" on it is not exactly a correct way to fetch selected value is it ?

driesvints:
A select isn't an input. I think you're looking for https://laravel.com/docs/5.8/dusk#assert-selected

Can you please post the entire test?

zoispag:
@driesvints consider it closed! The wrong assertion was indeed used in our test.

PR #935, opened 2021-10-12
Allow calling `assertValue()` on elements which don't support the `value` attribute

u01jmg3:
### Reasoning

Similar to [`assertInputValue()`](https://github.com/laravel/dusk/blob/6.x/src/Concerns/MakesAssertions.php#L375) and its use of [`inputValue()`](https://github.com/laravel/dusk/blob/6.x/src/Concerns/MakesAssertions.php#L410), I opted to create `elementValue()`, which for [a list of supported elements](https://www.w3schools.com/tags/att_value.asp) will retrieve the `value` attribute. If not it will fallback to calling [`getText()`](https://github.com/laravel/dusk/blob/6.x/src/Concerns/MakesAssertions.php#L416) on the element in the same way `inputValue()` currently works.

### Considerations

1. One scenario that is currently not possible before this PR, and continues not to be possible is a user wishing to use `assertValue()` against, for example, an `li` element and expecting to test against text between its tags. Because an `li` element supports the `value` attribute, using `assertValue()` cannot ever cover this scenario.
1. Although a `textarea` doesn't have a `value` attribute, its content is still retrieved using `$textarea->getAttribute('value')`. 
   - See https://github.com/laravel/dusk/issues/633#issuecomment-477286834 and PR https://github.com/laravel/dusk/pull/237.

### Why not use `inputValue()` and extend [the list of supported elements](https://github.com/laravel/dusk/blob/6.x/src/Concerns/MakesAssertions.php#L414)?

- This method calls `resolveForTyping()` but not all elements that support the `value` attribute allow typing.
- Although it would work, `inputValue()` expects a variable called `$field` yet it would be passed `$selector`. This to me is messy.

### Alternative Approaches

1. Add an additional assertion within `assertValue()` and `assertValueIsNot()` to check that someone is using these methods against an element that supports the `value` attribute. If not, throw an error.
1. Do nothing - this means users could inadvertently be calling `assertValue()` against an element such as a `div`:

   ```php
   // <div class="foo"></div>

   $browser->assertValue('.foo', ''); // true

   // <div class="foo">bar</div>

   $browser->assertValue('.foo', ''); // true
   ```

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -417,6 +417,32 @@ JS;
     }
 
     /**
+     * Get the value of the given selector.
+     *
+     * @param  string  $selector
+     * @return string
+     */
+    public function elementValue($selector)
+    {
+        $element = $this->resolver->findOrFail($selector);
+
+        $elementsSupportingTheValueAttribute = [
+            'button',
+            'input',
+            'li',
+            'meter',
+            'option',
+            'param',
+            'progress',
+            'textarea',
+        ];
+
+        return in_array($element->getTagName(), $elementsSupportingTheValueAttribute)
+                        ? $element->getAttribute('value')
+                        : $element->getText();
+    }
+
+    /**
      * Assert that the given input field is present.
      *
      * @param  string  $field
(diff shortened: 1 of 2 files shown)


driesvints:
> Again I agree but then I saw assertInputValue() supports a textarea which doesn't have a value attribute.

Hmm, that's a good remark. 

> happy to submit another PR that throws an error. However, it seems I've confused Taylor. Assuming you know what I'm trying to accomplish, should I try another PR?

I talked about that with Taylor and yes feel free to resubmit but we're gonna go with your proposal to only allow elements with a `value` attribute + textareas.

Issue #934, opened 2021-10-11
Use of `assertValue()` on elements which don't support the `value` attribute


u01jmg3:
- Dusk Version: 6.18.1
- Laravel Version: 6.20.35
- PHP Version: 7.4 / 8.1

### Description:

Recently I realised [`assertValue()`](https://laravel.com/docs/8.x/dusk#assert-value) can be used with any element, even those that don't support the `value` attribute.

```php
// <div class="foo">bar</div>

$browser->assertValue('.foo', ''); // Will always be true

$browser->assertValue('.foo', 'bar'); // Will always be false
```

Should there be a check that `assertValue()` is being called on an element that supports the `value` attribute?

These are the 7 elements which support the `value` attribute ([source](https://www.w3schools.com/tags/att_value.asp)):

- `button`
- `input`
- `li`
- `meter`
- `option`
- `progress`
- `param`

Happy to put together a PR if this is something you are interested in rectifying.

PR #936, opened 2021-10-12
#934 - throw an error if `assertValue()` is used with an element that does not support the `value` attribute

u01jmg3:
Throw an error if `assertValue()` / `assertValueIsNot()` are used with an element that does not support the `value` attribute.

Note: a `textarea` is also permitted even though it doesn't have a `value` attribute. This is because its content is still retrieved using `$textarea->getAttribute('value')`.

The list of elements that support the `value` attribute was sourced from: https://www.w3schools.com/tags/att_value.asp

---

(Previously closed PR: #935)

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -649,7 +649,15 @@ JS;
     {
         $fullSelector = $this->resolver->format($selector);
 
-        $actual = $this->resolver->findOrFail($selector)->getAttribute('value');
+        $element = $this->resolver->findOrFail($selector);
+        $tagName = $element->getTagName();
+
+        PHPUnit::assertTrue(
+            $tagName === 'textarea' || $this->supportsTheValueAttribute($tagName),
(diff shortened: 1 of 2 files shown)


lk77:
Why this change ?

this was working fine

lk77:
It's a fail safe, we could do without it, but with custom components, especially third party one, we are not sure that input event is properly triggered.

i don't really ask this to be reverted tho, i understand w3c compliance concerns

u01jmg3:
Thanks for the further explanation

I guess it depends what is more important. Supporting the use case you've described or preventing developers mistakenly calling `assertValue` on an element that doesn't support the `value` attribute and will therefore always have a value of `null` (`assertValue('')` is always `true`). 🤷🏻‍♂️

lk77:
to be fair, this behaviour has changed in vue3, it's modelValue now, to distinguish between native value,
i don't think it's worth fixing my specific use case.

u01jmg3:
Good change to make in order to distinguish

Glad you've found a middle ground ⛰️

Issue #69, opened 2017-01-27
Other selectors than css? xpath?


alex-LE:
Might it be possible to support different selectors than the cssSelector?
The webdriver also supports xpath selectors.
It would be great if functions like text() will support xpath also.

I've played a little bit with the Dusk code and added a parameter to the text() function and it seems to be very easy. But i'm not sure if this is the best way.
Dusk might also detect xpath by itself (by search for // at the beginning, for example).

What do you think?

alex-LE:
Ok a regex will not work, as xpath not only starts with //.
So maybye parameters to the functions like text() or passing a WebDriverBy object as selector? 
But this might also become difficult for functions like clickLink which uses the selector as a string.
I really want to help with that, but i don't know what the best idea may be.

@taylorotwell does such a feature makes sense?

noxify:
Hi,

i had the same problem.

My current workaround is to call the driver directly via:

```
$browser->driver->findElement( WebDriverBy::xpath("//table[@class='x-grid3-row-table']/tbody/tr/td/div/a[contains(text(),'$value')]") )
                    ->click();
```

noxify:
Or here a more "dynamic" way with the latest version (tested with the master branch).
(For details, please check: https://github.com/laravel/dusk/pull/119)

You can now extend the existing browser class, this allows us the following:

app\Testing\MyBrowser.php:

```
<?php

namespace App\Testing;

use Laravel\Dusk\Browser;

class MyBrowser extends Browser {

    /**
     * Create a browser instance.
     *
     * @param  \Facebook\WebDriver\Remote\RemoteWebDriver  $driver
     * @param  ElementResolver  $resolver
     * @return void
     */
    public function __construct($driver, $resolver = null)
    {
        $this->driver = $driver;

        $this->resolver = $resolver ?: new DynamicElementResolver($driver);
    }

    /**
     * Find an element by the given selector or return null.
     *
     * @param  string|\Facebook\WebDriver\WebDriverBy  $selector
     * @return \Facebook\WebDriver\Remote\RemoteWebElement|null
     */
    public function findBySelector($selector) {
        return $this->resolver->find($selector);
    }

}
```

app\Testing\DynamicElementResolver.php

```
<?php

namespace App\Testing;

use Laravel\Dusk\ElementResolver;
use Facebook\WebDriver\WebDriverBy;

class DynamicElementResolver extends ElementResolver
{  
    /**
     * Find an element by the given selector or throw an exception.
     *
[...]

PR #958, opened 2022-01-14
Add assertion: assertAttributeWithValue

tomosterlund:
This pull requests aims to make it possible to test the assertion, that there is an element with a given attribute, and a specific value for that attribute. The assertion should fail when:

1. There is no element with the specified attribute
2. When there is an element with the specified attribute, but it does not contain the specified value.

This will help developers in "marking" elements with attributes of their choice, instead of having to add ids or classes to target an element. For example, a developer could then create a `data-dusk` attribute on an element, and then target the element through `data-dusk` in their tests.

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -771,6 +771,23 @@ JS;
     }
 
     /**
+     * Assert that there is an element with the provided attribute and attribute-value.
+     *
+     * @param  string  $attributeName
+     * @param  string  $attributeValue
+     * @return $this
+     */
+    public function assertAttributeWithValue($attributeName, $attributeValue)
+    {
+        PHPUnit::assertTrue(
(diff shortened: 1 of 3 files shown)


taylorotwell:
I'm sorry I guess I'm just not following the need for this method so would have trouble maintaining it. Why can't you use `assertAttributeContains`?

tomosterlund:
Because it requires me to be able to first identify an element through its class or id selector, before I assert that it has an attribute with a certain value.

I tried out Dusk for the first time this week, and several times already I found myself adding classes and ids to elements, which they don't need, apart from being used in testing. So for me its a matter of separation of concerns. I don't want to start writing class attributes such as `class="c-button c-button--large _dusk-login_"`.

tomosterlund:
Also seems to me that others might wish being able to target elements based on non-CSS attributes:
https://github.com/laravel/dusk/issues/69

PR #967, opened 2022-02-24
[6.x] Allow to fill form fields targeted by their label

glaszig:
given the following markup

```html
<label for="testing">Laravel</label>
<input id="testing" type="text" />
```

one can now write a more semantic test like

```php
$browser->type("Laravel", "Dusk");
```

instead of having to target form fields by name or id

```php
$browser->type("testing", "Dusk");
```

there's method's like `assertSee()` which validate visible text on the page so, in line with that, it should be possible to fill fields using their labels.

### DIFF
--- a/src/ElementResolver.php
+++ b/src/ElementResolver.php
@@ -86,9 +86,17 @@ class ElementResolver
             return $element;
         }
 
-        return $this->firstOrFail([
-            "input[name='{$field}']", "textarea[name='{$field}']", $field,
-        ]);
+        $selectors = [
+            "input[name='{$field}']", "textarea[name='{$field}']", $field
+        ];
+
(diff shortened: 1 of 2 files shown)


driesvints:
I really believe you should just use an ID or name to target the input. I get that this looks good but the mechanism for targeting elements was designed to target attributes specifically. The way element values behave (like a label) is intended for them to be non-unique while name and ID are unique to target.

glaszig:
a field’s name attribute is not unique and targeting by name/id continues to work.

may only be me but coming from rails it irked me that i can `press()` and `clickLink()` by their “label” but cannot target fields that way. very unintuitive. 

in a follow up i planned to implement the same for select, radio, checkbox.

taylorotwell:

PR #980, opened 2022-06-04
Store source code and store console log on test fail

zsoltjanes:
I think it gives more flexibility for debugging our test cases when it fails. If we save the full DOM and the console messages it helps us debug easier the tests, screenshots tells nothing what would be the problem in my oppinion.


taylorotwell:

zsoltjanes:
:(

PR #1007, opened 2022-11-11
Allow custom chromedriver install path

underdpt:
Hi,

This PR allows to set a custom chromedriver install path. This woul allow users to install chromedriver outside vendor folder, and also enables the use of this package on phar apps (we cannot execute a chromedriver that's inside the phar file, so this allows the end user to install/update the chromedriver executable outside of the phar).

If no install path is configured, a backwards-compatible default one is used.

This PR uses a `dusk.php` config file, that could be used on the future for further dusk configuration (for example, package [nunomaduro/laravel-console-dusk](https://github.com/nunomaduro/laravel-console-dusk) makes use of a configuration file to setup the log and screenshots path, and also to enable not-headless run).


crynobone:
This is going to break all applications using Dusk.

`TestCase::buildChromeProcess()` gets called during `@beforeClass` where "Laravel" isn't ready yet alone `config`.

To demonstrate, A fresh installation with Dusk will now throw an error:

image

driesvints:
Thanks @crynobone, I've reverted this one. Is there some way we can add tests to cover that?

driesvints:
@underdpt looks like we can't make this happen in this way. We'll need to find another way if you want a custom path.

PR #1012, opened 2022-12-14
feat: add alternative tag for element selection

pataar:
This adds the alternative `data-dusk="tag"` attribute. As described in: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes: attributes that are not part of the spec should be prefixed with `data-`. 

Currently TSX does not allow any attributes that aren't defined in the spec, unless they have a data prefix. So without an alternative attribute to `dusk=`, TSX users won't be able to use Dusk as documented.

### DIFF
--- a/src/ElementResolver.php
+++ b/src/ElementResolver.php
@@ -410,7 +410,11 @@ class ElementResolver
         );
 
         if (Str::startsWith($selector, '@') && $selector === $originalSelector) {
-            $selector = '[dusk="'.explode('@', $selector)[1].'"]';
+            $elementPieces = explode('@', $selector);
+            $selector = $this->prefix . ' [dusk="'.$elementPieces[1].'"],';
+            $selector .= $this->prefix . ' [data-dusk="'.$elementPieces[1].'"]';
+
+            return trim($selector);
         }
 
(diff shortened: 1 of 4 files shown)


driesvints:
We just reverted this because it was breaking Nova's test suite.

crynobone:
Hi, 

I do believe this is a great addition to Dusk and Nova should use `data-dusk` on the next major version however as said earlier we need to revert this for now. 

This PR should work great in some scenario however it browse any interaction with elements where we need to use `ElementResolver::findOrFail()` etc. A simple example is `$browser->keys('@notes', 'Tests');` now failed.

pataar:
> We just reverted this because it was breaking Nova's test suite.

@driesvints Any chance in bringing this back in the future?

driesvints:
@pataar mior answered this above your reply.

pataar:
> @pataar mior answered this above your reply.

@driesvints Alright, thanks!

pataar:
This is now added by Taylor: https://github.com/laravel/dusk/commit/cf04717664f80204567ad3077ea7484a0be16497

PR #1019, opened 2023-01-17
Include verification of the value

shealavington:
isset is not enough to consider the state of the options.  

With the previous solution, even if DUSK_HEADLESS_DISABLED=false, the result was truthy because the value was set.

The proposed solution also checks the value.

```
<?php

// Outputs  'FALSE || 0'
echo isset($DUSK_HEADLESS_DISABLED) ? '_SET' : '_UNSET';
	
// Outputs  'TRUE || 0'
$DUSK_HEADLESS_DISABLED = true;
echo isset($DUSK_HEADLESS_DISABLED) ? '_SET' : '_UNSET';

// Outputs 'TRUE || 1'
$DUSK_HEADLESS_DISABLED = false;
echo isset($DUSK_HEADLESS_DISABLED) ? '_SET' : '_UNSET';
```

Notice that last example, although the value is false, it would be considered truth with the existing implementation.

### DIFF
--- a/stubs/DuskTestCase.stub
+++ b/stubs/DuskTestCase.stub
@@ -51,8 +51,8 @@ abstract class DuskTestCase extends BaseTestCase
      */
     protected function hasHeadlessDisabled(): bool
     {
-        return isset($_SERVER['DUSK_HEADLESS_DISABLED']) ||
-               isset($_ENV['DUSK_HEADLESS_DISABLED']);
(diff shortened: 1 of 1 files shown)


shealavington:
The failing test is not an issue I can solve.

image

taylorotwell:
This code has a syntax error for unmatched parens, which tells me it hasn't been run or tested. 😅

PR #1020, opened 2023-01-18
Include verification of the value (Fixed Brackets)

shealavington:
**Using `isset` is not a flexible/good enough check to consider the state of the options.**  

With the currently implemented solution, even if `DUSK_HEADLESS_DISABLED=false`, the function returns truthy because the value "is set".

My proposed solution also checks the value.

Example showing that the current implementation is returning truthy.
```php
/*
 * Currently implemented code.
 * Outputs 'TRUE || 1' 
 * Despite being explicitly set to FALSE, it would return true.
 */
$DUSK_HEADLESS_DISABLED = false;
echo isset($DUSK_HEADLESS_DISABLED) ? 'true' : 'false';

/*
 * Suggested change
 * Outputs 'FALSE || 0' 
 * Being explicitly set to FALSE, it returns false.
 */
$DUSK_HEADLESS_DISABLED = false;
echo (isset($DUSK_HEADLESS_DISABLED) && $DUSK_HEADLESS_DISABLED) ? 'true' : 'false';
```

> The code was tested on a local install of Laravel using Dusk. 
> Change suggestion made using Github's interface to make my contribution back to Dusk easier.

_Please suggest feedback before closing if you disagree with the solution proposed._

### DIFF
--- a/stubs/DuskTestCase.stub
+++ b/stubs/DuskTestCase.stub
@@ -52,7 +52,7 @@ abstract class DuskTestCase extends BaseTestCase
     protected function hasHeadlessDisabled(): bool
     {
         return (isset($_SERVER['DUSK_HEADLESS_DISABLED']) && $_SERVER['DUSK_HEADLESS_DISABLED']) ||
-               (isset($_ENV['DUSK_HEADLESS_DISABLED']) && $_ENV['DUSK_HEADLESS_DISABLED'];
+               (isset($_ENV['DUSK_HEADLESS_DISABLED']) && $_ENV['DUSK_HEADLESS_DISABLED']);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


shealavington:
A little disappointed that PRs are just closed instead of requesting feedback or considering the actual issue.
I used the GitHub UI to suggest the changes; mistakes happen.

I suppose something must be wrong as tests failed again, I'll look more if I get time, but issue still stands.

shealavington:
I've had a look, I genuinely don't know how my suggestion would cause that test to fail.
I'd love some feedback or thoughts rather than just closing the PR...

taylorotwell:
I'm not positive you're running this code? If you put `EXAMPLE=false` in your `.env` file, the $_ENV var will be a string containing `false`, which is "truthy".

If PRs have syntax errors they are closed - these tools are used by thousands of folks. Sorry.

shealavington:
Fair enough, so this is expected behaviour. Closure makes sense then 👍

PR #1024, opened 2023-02-10
[7.x] Fix tests

driesvints:


driesvints:
Ping @crynobone

src/Console/DuskCommand.php
```diff
@@ -282,7 +283,11 @@ protected function writeConfiguration()
     {
         if (! file_exists($file = base_path('phpunit.dusk.xml')) &&
             ! file_exists(base_path('phpunit.dusk.xml.dist'))) {
-            copy(realpath(__DIR__.'/../../stubs/phpunit.xml'), $file);
+            if (version_compare(Application::VERSION, '10.0.0') >= 0) {
```
crynobone, line 286:
```suggestion
            if (version_compare(Application::VERSION, '10.0', '>=')) {
```

crynobone: But this doesn't guarantee PHPUnit 10 is installed, maybe use `version_compare()` against phpunit instead?

taylorotwell:
@driesvints please mark as ready for review when feedback addressed.

src/Console/DuskCommand.php
```diff
@@ -4,6 +4,7 @@
 
 use Dotenv\Dotenv;
 use Illuminate\Console\Command;
+use Illuminate\Foundation\Application;
 use Illuminate\Support\Str;
```
crynobone, line 8:
```suggestion
use Illuminate\Support\Str;
use PHPUnit\Runner\Version;
```

src/Console/DuskCommand.php
crynobone, line 286:
```suggestion
            if (version_compare(Version::id(), '10.0', '>=')) {
```

driesvints:
Ready again.

PR #1027, opened 2023-02-21
[7.x] Use `--headless=new`

SjorsO:
As of Chrome version 109, there is a new and improved headless mode that can be enabled with `--headless=new`. It is described in more detail in this [Selenium blog post](https://www.selenium.dev/blog/2023/headless-is-going-away/) and this [StackOverflow comment](https://stackoverflow.com/questions/45631715/downloading-with-chrome-headless-and-selenium/73840130#73840130).

The `--headless=new` flag also fixes a breaking change in Chromedriver where setting a custom download directory stopped working. I was using the code below to set a custom download directory:

```php
$options->setExperimentalOption('prefs', [
    'download.default_directory' => storage_path('app/dusk-downloads'),
]);
```

This stopped working recently and caused my tests to fail, but setting `--headless=new` fixed it.

### DIFF
--- a/stubs/DuskTestCase.stub
+++ b/stubs/DuskTestCase.stub
@@ -34,7 +34,7 @@ abstract class DuskTestCase extends BaseTestCase
         ])->unless($this->hasHeadlessDisabled(), function (Collection $items) {
             return $items->merge([
                 '--disable-gpu',
-                '--headless',
+                '--headless=new',
             ]);
         })->all());
 


crynobone:
@SjorsO Would there be any issue if we run `--headless=new` on Chrome version lower than 109? Locally you may be running the latest version but CI environment may still be using an older version.

SjorsO:
> @SjorsO Would there be any issue if we run `--headless=new` on Chrome version lower than 109? Locally you may be running the latest version but CI environment may still be using an older version.

I'm not sure, my Chrome is already past version 109 so I can't test it. But since this change only affects new installs of Dusk, and since Chrome auto-updates, I don't think it will be a problem.

PR #1029, opened 2023-03-17
Fixed timeout error message that contain percent characters when calling waitForTextIn

voidgraphics:
Hello, I have found a bug with the timeout error message when calling `waitForTextIn` with text that contains a percent character.

For example: 

A failing call to `waitForTextIn('foo', 'Discount: 20%')` should display the following error message: 

```
Waited 1 seconds for text "Discount: 20%" in selector foo
```

But it actually shows 

```
ArgumentCountError: 3 arguments are required, 2 given
```

Which is caused by the `sprintf` call to format the error message getting confused by the extra % signs in the message.

I believe this has already been fixed for `waitForText`, as evidenced by the `test_wait_for_text_failure_message_containing_a_percent_character` test, but the bug remains for `waitForTextIn`.

I tried using the `formatTimeOutMessage` method which was used to fix this bug on `waitForText`, but it formats the error message differently and is not easily compatible with the error message format that's expected from the `waitForTextIn` method.

So my fix is simpler, and I've just added a `str_replace` call. It could be a good idea to refactor the `formatTimeOutMessage`, and make the error messages more uniform in the future, but that could be considered a breaking change I think, so I did not venture there.

### DIFF
--- a/src/Concerns/WaitsForElements.php
+++ b/src/Concerns/WaitsForElements.php
@@ -123,7 +123,7 @@ trait WaitsForElements
      */
     public function waitForTextIn($selector, $text, $seconds = null)
     {
-        $message = 'Waited %s seconds for text "'.$text.'" in selector '.$selector;
+        $message = 'Waited %s seconds for text "'.str_replace('%', '%%', $text).'" in selector '.$selector;
 
         return $this->waitUsing($seconds, 100, function () use ($selector, $text) {
             return $this->assertSeeIn($selector, $text);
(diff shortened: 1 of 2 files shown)


taylorotwell:
Mark as ready for review when I should take a look again please.

driesvints:
@voidgraphics are you still working on this?

voidgraphics:
Sorry, forgot about this. I'll make some time for it this weekend!

voidgraphics:
I have refactored the fix a little bit to avoid duplication.

> Wouldn't this apply to other messages as well? We would want to fix it everywhere?

A fix was already in place for the other messages, `waitForTextIn` was the outlier and did not use the fix because its error message is formatted differently. Some tests were missing for these other methods though, so I added them.

PR #1032, opened 2023-04-03
Add support for native scroll into view options

aedart:
This expands the work from @u01jmg3 (#765), by adding support for the native "scrollIntoViewOptions", as described by [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView).
Furthermore, PR also adds a shortcut method for scrolling an element "instantly" into view.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -386,10 +386,10 @@ class Browser
     {
         $selector = addslashes($this->resolver->format($selector));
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
It doesn't seem like this actually fixes any bugs, so will leave as is for now.

PR #1033, opened 2023-04-09
Show error if the test class name doesn't have the keyword Test appended.

Abhi0725:
This is just a QOL improvement. I have found myself in situation when I use the dusk make command as following:

`php artisan dusk:make Dashboard`

In this case I want to make a Dashboard test but forgot to append the name with keyword "Test". As the dusk class name isn't DashboardTest, the tests just simply skipped silently. 

I wanted to make it this way where the command asks the input again but couldn't figure this out without overriding the `getNameInput()` method (and I feel that's overkill for this small improvement). 

Hence this PR just simply throws an error message if the "Test" keyword isn't appended. Let me know if it should be done differently.

### DIFF
--- a/src/Console/MakeCommand.php
+++ b/src/Console/MakeCommand.php
@@ -71,4 +71,15 @@ class MakeCommand extends GeneratorCommand
     {
         return 'Tests';
     }
+
+    public function handle()
+    {
+        $filename = $this->argument('name');
+
+        if(! Str::endsWith($filename, 'Test')) {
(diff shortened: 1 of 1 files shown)


Abhi0725:
@taylorotwell this will improve DX which laravel always cares for. In fact this PR (https://github.com/laravel/dusk) does something similar.

I and some of my friends have experienced this issue many times when we forget to append  the keyword "Test" in class name and the tests simply passes because it just ignores those classes. I think there will be others too who might face the same.

PR #1034, opened 2023-04-13
Allow `dusk` attribute selectors to be chained

JayBizzle:
I have come across a few situations when writing Dusk tests where I use the `dusk` attribute selector on a wrapper `div`, then want to select elements _within_ that `div`...

e.g.
```html
<div dusk="all-items">
    <div>Item 1</div>
    <div>Item 2</div>
    <div>Item 3</div>
    <div>Item 4</div>
</div>
````

And I was hoping I could do something like this...

```php
$items = $browser->elements('@all-items > div');

foreach($items as $item){
    // assert something
}
```

...but this doesn't work because the selector gets compiled to `[dusk="all-items > div"]`

This PR makes this selector get compiled to `[dusk="all-items"] > div`

### Bonus
Also allows multiple `dusk` attribute selectors to work as expected

This selector...
```
@item-list > div:nth-child(2) @submit-button
```

becomes...
```
[dusk="item-list"] > div:nth-child(2) [dusk="submit-button"]
```

### DIFF
--- a/src/ElementResolver.php
+++ b/src/ElementResolver.php
@@ -410,7 +410,7 @@ class ElementResolver
         );
 
         if (Str::startsWith($selector, '@') && $selector === $originalSelector) {
-            $selector = '['.Dusk::$selectorHtmlAttribute.'="'.explode('@', $selector)[1].'"]';
+            $selector = preg_replace('/@(\S+)/', '['.Dusk::$selectorHtmlAttribute.'="$1"]', $selector);
         }
 
         return trim($this->prefix.' '.$selector);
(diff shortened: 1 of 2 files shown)


patrickomeara:
This has broken dusk selectors with spaces.

JayBizzle:
> This has broken dusk selectors with spaces.

Thanks for reporting, will submit a fix when I get the the office in a couple of hours 👍🏻

patrickomeara:
@JayBizzle No troubles. In my opinion the functionality that you have added here is worth the breaking change. I'd be interested to see how wide spread the space issue is. Swapping my spaces out won't be too much work, it's mainly for selecting submenus in react components. A breaking change nonetheless. Thanks for the quick response.

JayBizzle:
Yes, that is the problem I have hit with the solution. There doesn't seem to be a way around that unfortunately.

See #1035 to continue discussion 👍

driesvints:
@patrickomeara seems Taylor has decided to continue with this PR, sorry.

PR #1035, opened 2023-04-19
Allow `dusk` selectors to contain spaces

JayBizzle:
As reported by @patrickomeara, PR #1034 broke tests for anyone using `dusk` selectors that contained a space character.

This PR fixes that.

**NOTE**
I have limited the characters that can be matched in a `dusk` selector to alphanumeric chars plus `:`, `_`, `-`, `[:space:]` because we can't allow chars like `>` or `.` because that would then start matching chained selectors like...

```
@dusk selector .class
@dusk selector > div
```

The biggest issue with this fix though is you will not be able to do a selector like this...

```
@dusk selector div
```

...as the regex doesn't know which space to stop matching on. This will compile to...

```
dusk=["dusk selector div"]
```

rather than the desired
```
dusk=["dusk selector"] div
```

So, I see 3 options here...

1. Revert #1034
2. Keep as-is and note in the docs that `dusk` selectors cannot contain spaces
3. Merge this PR which adds complexity and fixes the space issue, but doesn't allow all types of CSS selector chaining

My preference would be option 2, but then I would say that as it suits my requirements, but I was a little perplexed when I couldn't chain selectors, seemed like an obvious thing to allow.

Options 3 (this PR) worries me as it feels like we may end up chasing our tails.

Thanks

### DIFF
--- a/src/ElementResolver.php
+++ b/src/ElementResolver.php
@@ -410,7 +410,13 @@ class ElementResolver
         );
 
         if (Str::startsWith($selector, '@') && $selector === $originalSelector) {
-            $selector = preg_replace('/@(\S+)/', '['.Dusk::$selectorHtmlAttribute.'="$1"]', $selector);
+            $selector = $selector.' ';
+            $selector = trim(
+                preg_replace(
+                    '/@([a-zA-Z0-9\:\_\-\ ]+)(?=\s)/',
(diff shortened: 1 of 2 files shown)


taylorotwell:
Going to lead it ride as-is for now.

ChristianOliveiraSilva:
I am facing problems with space in the dusk selector, have we taken any initiative to implement a solution to this problem?

PR #1036, opened 2023-06-29
Use the default printer from Collision

patrickomeara:
I would like some feedback on some speed bumps I have hit.

My main objective is getting a nicer console output from the dusk command and keep some consistency with `php artisan test`. The additional benefit is getting quicker feedback from CI environments that print line by line.

PHPUnit 10 dropped support for printers and instead introduced subscribers. @nunomaduro's `EnsurePrinterIsRegisteredSubscriber` class is a subscriber. However this class is marked as `@internal` so I'm unsure if we should use it in Dusk or follow best practices and break it out into it's own package that both Collision and Dusk, and others can use. Any feedback would be appreciated.

I can confirm that this is working in collision 7/phpunit 10.

image

I've added the collision constraint as `^6.0|^7.0` for backwards compatibility. Although, the printer won't work in PHPUnit 9.

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "ext-zip": "*",
         "php-webdriver/webdriver": "^1.9.0",
         "nesbot/carbon": "^2.0",
-        "nunomaduro/collision": "^7.7",
+        "nunomaduro/collision": "^6.0|^7.0",
         "illuminate/console": "^9.0|^10.0",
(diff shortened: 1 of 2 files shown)


nunomaduro:
Thank you for this pull request. I've adjusted the code here: https://github.com/laravel/dusk/pull/1038

PR #1041, opened 2023-07-19
[8.x] New Chromedriver downloads

driesvints:
This updates Dusk to the new urls from Google for getting the latest version of the chromedrivers as well as their download links for each OS. Chrome changed this to a JSON payload instead: https://chromedriver.chromium.org/home

> Starting with M115 the latest Chrome + ChromeDriver releases per release channel (Stable, Beta, Dev, Canary) are available at[ the Chrome for Testing availability dashboard](https://googlechromelabs.github.io/chrome-for-testing/). For automated version downloading one can use the convenient [JSON endpoints](https://googlechromelabs.github.io/chrome-for-testing/last-known-good-versions-with-downloads.json).

Here's a list of all endpoints: https://github.com/GoogleChromeLabs/chrome-for-testing#json-api-endpoints

I'm using the `last-known-good-versions-with-downloads.json` one to grab the latest stable version one and the `latest-versions-per-milestone-with-downloads.json` to easily match a requested milestone (e.g. 115) with their download url.

I've also changed the slug and OS matching to the new naming conventions. And also added support for Windows 64 bit chromedrivers (I can take out that part if you want).

To upgrade, people will need to install at least Chrome 115 and re-run `php artisan dusk:chrome-driver`.

I have not yet tested this out on a Laravel app.

PR #1042, opened 2023-07-20
[7.x] Fix ChromeDriverCommand to use Chrome 115 as new minimum

driesvints:
This is the same PR as https://github.com/laravel/dusk/pull/1041 but no windows 64 support (I don't think it's really needed) and it doesn't touches the chromeprocess so it's backwards compatible with the binaries developers already have installed. So it should be safe to do the upgrade on v7 of Dusk.

The only thing I still did was adopt the new urls and json payload Chrome sends so we still need to have 115 as a minimum version. This is the lowest version returned by https://googlechromelabs.github.io/chrome-for-testing/latest-versions-per-milestone-with-downloads.json that has chromedriver downloads.

### DIFF
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -100,8 +100,8 @@ class ChromeDriverCommand extends Command
 
         $milestone = (int) $version;
 
-        if ($milestone < 115) {
-            throw new Exception('Dusk v7 requires Chrome 115 or above.');
+        if ($milestone < 113) {
+            throw new Exception('Dusk v7 requires Chrome 113 or above.');
         }
 
         $all = $this->option('all');
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -100,8 +100,8 @@ class ChromeDriverCommand extends Command
 
         $milestone = (int) $version;
 
-        if ($milestone < 113) {
-            throw new Exception('Dusk v7 requires Chrome 113 or above.');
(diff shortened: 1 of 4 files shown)


glaszig:
> need to have 115 as a minimum version

the older scheme of acquiring chromedriver < 115 still works. there could be support for 2 behaviors in the code. one downloader for < 115 and another going forward.

driesvints:
> the older scheme of acquiring chromedriver < 115 still works. there could be support for 2 behaviors in the code. one downloader for < 115 and another going forward.

I get that and tried it but it became such a mess. So made the decision to just support 115 going forward... anyone who already has Dusk installed will still be able to use the binaries they're using. But going forward new Dusk installs will require Chrome 115 as a new minimum. Like you say, you can't get an older version of Chrome anyway.

src/Console/ChromeDriverCommand.php
```diff
@@ -211,7 +178,12 @@ protected function latestVersion()
             $streamOptions['http'] = ['proxy' => $this->option('proxy'), 'request_fulluri' => true];
         }
 
-        return trim(file_get_contents($this->latestVersionUrl, false, stream_context_create($streamOptions)));
+        $versions = json_decode(file_get_contents(
+            $this->latestVersionUrl, false, stream_context_create($streamOptions)
+        ), true);
```
crynobone, line 183:
```suggestion
        $versions = json_decode($this->getUrl($this->latestVersionUrl), true);
```

taylorotwell:
Merged Mior's PR.

PR #1043, opened 2023-07-22
[7.x] Fix ChromeDriverCommand for ChromeDriver 115+

crynobone:
fixes #1040

PR #1052, opened 2023-08-03
[7.x] Prevent interaction with `Http::preventStrayRequests`

joshbonnick:
This PR replaces the use of the `Http` facade with direct use of the Guzzle client allowing the request to be made regardless of stray requests being prevented.

Since #1043 the `getUrl` method for the `dusk:chrome-driver` command has used the `Http` facade to get the URL for the latest download URL. This has the unintended consequence of throwing an exception if stray requests are being prevented. Previously the URL was obtained using `file_get_contents` so it would still allow the command to successfully execute when requests are being prevented.

The Guzzle client is also directly used in the `download` method already.

### DIFF
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -6,7 +6,6 @@ use Exception;
 use GuzzleHttp\Client;
 use GuzzleHttp\Psr7\Utils;
 use Illuminate\Console\Command;
-use Illuminate\Support\Facades\Http;
 use Illuminate\Support\Str;
 use Laravel\Dusk\OperatingSystem;
 use Symfony\Component\Process\Process;
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -307,6 +307,7 @@ class ChromeDriverCommand extends Command
      * Get the contents of a URL using the 'proxy' and 'ssl-no-verify' command options.
      *
      * @return string
+     *
      * @throws Exception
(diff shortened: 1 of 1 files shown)


src/Console/ChromeDriverCommand.php
```diff
@@ -311,10 +310,10 @@ protected function resolveChromeDriverDownloadUrl(string $version, string $os)
      */
     protected function getUrl(string $url)
     {
-        return Http::withOptions(array_merge([
+        return (new Client())->get($url, array_merge([
             'verify' => $this->option('ssl-no-verify') === false,
         ], array_filter([
             'proxy' => $this->option('proxy'),
-        ])))->get($url)->body();
+        ])))->getBody();
```
crynobone, line 317: There two issue with these changes if we want to replace HTTP Client:

* We need to handle none 2xx response and throw exception
* Response may return a stream instead of string, it best to cast the return value to string.

Reference: https://github.com/laravel/framework/blob/d3586ebfd88d1ac62035289d1172b7978cc8e429/src/Illuminate/Http/Client/PendingRequest.php#L872-L933

crynobone: For example: https://github.com/orchestral/dusk-updater/blob/f407d18d7134c4d29e387f00909222f12a230670/src/helpers.php#L55-L67

joshbonnick: @crynobone changes added as requested. The exception handling is the same as the download methods exception handling.

PR #1053, opened 2023-08-28
[7.x] Keyboard Support Improvements

crynobone:
### Before

```php
$key = PHP_OS_FAMILY === 'Darwin' ? \Facebook\WebDriver\WebDriverKeys::COMMAND : \Facebook\WebDriver\WebDriverKeys::CONTROL;

$this->browse(function (Browser $browser) use ($key) {
    $browser
        ->visit('/first')
        // ...
        ->tap(function (Browser $browser) use ($key) {
            $browser->driver->getKeyboard()->pressKey($key);
        })
        ->click('@link.to.second')
        ->tap(function (Browser $browser) use ($key) {
            $browser->driver->getKeyboard()->releaseKey($key);
        })
        // ...
    ;
});
```

### After

```php
$key = PHP_OS_FAMILY === 'Darwin' ? \Facebook\WebDriver\WebDriverKeys::COMMAND : \Facebook\WebDriver\WebDriverKeys::CONTROL;

$this->browse(function (Browser $browser) use ($key) {
    $browser
        ->visit('/first')
        // ...
        ->usesKeyboard(function (Keyboard $keyboard) use ($key) {
            $keyboard->press($key);
            $keyboard->browser->click('@link.to.second');
            $keyboard->release($key);
        }) 
        // or alternatively
        ->controlClick('@link.to.second')
        // ...
    ;
});
```

### DIFF
--- a/src/Concerns/InteractsWithMouse.php
+++ b/src/Concerns/InteractsWithMouse.php
@@ -5,9 +5,9 @@ namespace Laravel\Dusk\Concerns;
 use Facebook\WebDriver\Exception\ElementClickInterceptedException;
 use Facebook\WebDriver\Exception\NoSuchElementException;
 use Facebook\WebDriver\Interactions\WebDriverActions;
-use Facebook\WebDriver\Remote\RemoteKeyboard;
 use Facebook\WebDriver\WebDriverBy;
 use Facebook\WebDriver\WebDriverKeys;
+use Laravel\Dusk\Keyboard;
 use Laravel\Dusk\OperatingSystem;
 
 trait InteractsWithMouse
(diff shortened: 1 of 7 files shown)


src/Concerns/InteractsWithMouse.php
```diff
@@ -153,6 +156,23 @@ public function rightClick($selector = null)
...
+    /**
+     * Control click the element at the given selector.
+     *
+     * @param  string|null  $selector
+     * @return $this
+     */
+    public function controlClick($selector = null)
+    {
+        return $this->withKeyboard(function (Keyboard $keyboard) use ($selector) {
+            $key = OperatingSystem::onMac() ? WebDriverKeys::META : WebDriverKeys::CONTROL;
```
taylorotwell, line 168: I don't quite follow this part. On my macOS keyboard I have a control key **and** a "meta" / command key. Why do we use the meta key for Mac here if some Mac keyboards have an actual control key?

crynobone: `COMMAND` and `META` represent the same keystroke in Mac : https://github.com/php-webdriver/php-webdriver/blob/a1578689290055586f1ee51eaf0ec9d52895bb6d/lib/WebDriverKeys.php#L90

The primary issue here is that Windows uses Control+click while Mac uses Command+click to trigger open in new tab. Similar issue can occur for action combo such as copy text etc.

PR #1059, opened 2023-09-21
Add a see number of elements on page assertion helper

ziadoz:
This PR adds a new helper to assert the number of matching elements on the page:

```
$browser->assertNumberOfElements('foo', 1);
$browser->assertNumberOfElements('foo', '=', 1);
$browser->assertNumberOfElements('foo', '>', 1);
$browser->assertNumberOfElements('foo', '>=', 1);
$browser->assertNumberOfElements('foo', '<', 1);
$browser->assertNumberOfElements('foo', '<=', 1);
```

In my case I'm converting a Codeception testsuite over to Dusk, and was looking for an equivalent to its `seeNumberOfElements()` method so I can test an infinite scroll works. 

With this new method you'd be able to write a test like this:

```
$browser->loginAs(User::factory()->create())
    ->visitRoute('search')
    ->waitUntilEnabled('[type="submit"]')
    ->press('Search')
    ->assertNumberOfElements('table.results tbody tr', '=', 40)
    ->scrollIntoView('.results-bottom')
    ->pause(1000) // Wait for records to load...
    ->assertNumberOfElements('table.results tbody tr', '>' 40);
```

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -5,6 +5,7 @@ namespace Laravel\Dusk\Concerns;
 use Facebook\WebDriver\Exception\NoSuchElementException;
 use Facebook\WebDriver\Remote\RemoteWebElement;
 use Illuminate\Support\Str;
+use InvalidArgumentException;
 use PHPUnit\Framework\Assert as PHPUnit;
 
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #1070, opened 2023-11-14
Limiting length of getCallerName return value to avoid issue when storing logs and screenshots

lk77:
Hello,

I have some dusk tests that fails when there is console logs, because the filename is too long

```
ErrorException : file_put_contents(/var/www/html/tests/Browser/console/Tests_Browser_Pages_SomeFolder_MoreFolders_MyFolder_MyTestClass_testFunction-0.log): Failed to open stream: File name too long
```

it happens when the test class is in a deep subfolder, in my case i have 5 folders, and also the name of the test itself is pretty long, 

i propose to limit the number of characters of the return value of the ProvidesBrowser::getCallerName function

```php
/**
 * Get the browser caller name.
 *
 * @return string
 */
protected function getCallerName()
{
    $name = version_compare(Version::id(), '10', '>=')
        ? $this->name()
        : $this->getName(false); // @phpstan-ignore-line

    return str_replace('\\', '_', substr(get_class($this), 0, 70)).'_'.substr($name, 0, 70);
}
```
both the class name and the test name will be limited to 70 characters

i added a test, to verify that the test name is correctly truncated,

to test that the class name is correctly truncated would require the creation of another test class with a very long name, 

i don't know if it's something you want to do.

### DIFF
--- a/src/Concerns/ProvidesBrowser.php
+++ b/src/Concerns/ProvidesBrowser.php
@@ -232,7 +232,7 @@ trait ProvidesBrowser
             ? $this->name()
             : $this->getName(false); // @phpstan-ignore-line
 
-        return str_replace('\\', '_', get_class($this)).'_'.$name;
+        return str_replace('\\', '_', substr(get_class($this), 0, 70)).'_'.substr($name, 0, 70);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ziadoz:
@lk77 I'm working on a [PR](https://github.com/laravel/dusk/pull/1084) that adds data sets to filenames for tests using data providers. Do you have any idea what the maximum file length is? And what operating system this issue occurs on? I just want to make sure I'm not re-introducing this bug.

lk77:
@lk77 it's 255, but for the whole path, that's why i limited to filename to 140 to be sure there is enough for the directory

PR #1077, opened 2024-01-07
Drop PHPUnit 9 and limits Laravel supported versions to 10 and 11

crynobone:


driesvints (approved):
We'll need a new major version for this one.

PR #1084, opened 2024-02-08
Add PHPUnit data provider set name to output filenames

ziadoz:
If there's a failure in a Dusk test case that uses PHPUnit data providers, then only a screenshot of the last data set that failed will be available in the `screenshots` directory. This is because without the data set in the filename, each new screenshot overwrites the previous one.

This PR adds the data set, where applicable, to the filenames of screenshots and console output:

With no data provider:
```
failure-Tests_Browser_FooTest_test_without_data-0.png
```

With unnamed data set (`0`, `1`): 
```
failure-Tests_Browser_FooTest_test_with_data_0-0.png
failure-Tests_Browser_FooTest_test_with_data_1-0.png
```

With named data set (`foo`, `bar`):
```
failure-Tests_Browser_FooTest_test_with_named_data_foo-0.png
failure-Tests_Browser_FooTest_test_with_named_data_bar-0.png
```

I've retained the existing [140 character filename limit](https://github.com/laravel/dusk/pull/1070) by trimming it off the left hand side of the filename, so namespaces are truncated.

### DIFF
--- a/tests/Unit/ProvidesBrowserTest.php
+++ b/tests/Unit/ProvidesBrowserTest.php
@@ -23,7 +23,7 @@ class ProvidesBrowserTest extends TestCase
     {
         $browser = m::mock(stdClass::class);
         $browser->shouldReceive('screenshot')->with(
-            'failure-Laravel_Dusk_Tests_Unit_ProvidesBrowserTest_test_capture_failures_for-0'
+            'failure-Laravel_Dusk_Tests_Unit_ProvidesBrowserTest_test_capture_failures_for_foo-0'
         );
         $browsers = collect([$browser]);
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can we make the PR not change the names of tests that use no data set?

tests/Unit/ProvidesBrowserTest.php
```diff
@@ -31,34 +28,51 @@ public function test_capture_failures_for()
...
-            'Laravel_Dusk_Tests_Unit_ProvidesBrowserTest_test_truncate_test_name_where_that_name_is_too_long_and_might_cause_is-0'
+            'Dusk_Tests_Unit_ProvidesBrowserTest_test_truncate_test_name_where_that_name_is_really_really_really_too_long_and_might_cause_issues_data_foo-0'
         );
         $browsers = collect([$browser]);
 
         $this->storeConsoleLogsFor($browsers);
     }
 
-    public static function testData()
+    public static function data()
```
ziadoz, line 72: I took the `test` prefix off of this because PHPUnit is running the method as a test.

PR #1088, opened 2024-02-27
New Assertions: assertPathEndsWith and assertPathContains

shawnhooper:
Adds two new assertions to go with the existing assertPathBeginsWith:

- assertPathEndsWith
- assertPathContains

### DIFF
--- a/src/Concerns/MakesUrlAssertions.php
+++ b/src/Concerns/MakesUrlAssertions.php
@@ -169,7 +169,7 @@ trait MakesUrlAssertions
     }
 
     /**
-     * Assert that the current URL path begins with the given path.
+     * Assert that the current URL path ends with the given path.
      *
      * @param  string  $path
      * @return $this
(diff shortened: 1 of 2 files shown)


taylorotwell:
I would maybe use Str::contains for the contains one.

shawnhooper:
Hi @taylorotwell, thanks for the feedback.  

Looking at the other assertions in the `MakesUrlAssertions` trait, and they all call assertions from the PHPUnit framework.  I think using `PHPUnit::assertStringContainsString()` is more consistent, and cleaner than checking Str::contains and then trying to throw a PHPUnit assertion error if it doesn't.

u01jmg3:
@driesvints: any chance we could get this PR progressed? Would be great to see `assertPathEndsWith()` in Dusk

driesvints:
Taylor doesn't sees PR's in draft state. If a new review is needed, the PR needs to be marked as such. I'll do this for now.

u01jmg3:
Thanks @driesvints - I can't change the state of a draft PR and I'm not sure it was clear to @shawnhooper that they weren't going to get a reply. Either way, this PR will get looked at now 👍🏻

shawnhooper:
@u01jmg3 Correct. I assumed Taylor has been busy and would get to it eventually.  First time contributing into the official packages. :)

u01jmg3:
@driesvints - I see this was merged into `7.x` and with today's `v8.1.2` release, this PR hasn't made it in. Should this PR have gone into the `8.x` branch? Perhaps because this PR was created back in February.

Shall I raise a new PR to add these changes to `8.x` or can you fix it?

driesvints:
This should indeed have gone into 8.x since 7.x is no longer maintained. If you could send in a new PR I'll merge and tag it. Thanks!

PR #1094, opened 2024-03-23
Support Dusk Selectors in `screenshotElement`

rabrowne85:
in the previous PR #1093 the use of Dusk selectors were not supported. This formats the selector as per other methods thereby supporting the use of Dusk selectors.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -449,6 +449,30 @@ class Browser
     }
 
     /**
+     * Take a screenshot of a specific element and store it with the given name.
+     *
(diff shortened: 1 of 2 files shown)


u01jmg3:
You need to fix the tests `test_element_screenshot` and `test_element_screenshot_in_subdirectory` due to your change.

rabrowne85:
Thanks @u01jmg3.

Tests have been revised to suit the corrections to the method. I also renamed the tests to follow the same naming of the method.

PR #1098, opened 2024-04-16
Move `hasHeadlessDisabled()` and `shouldStartMaximized()` to `Laravel\Dusk\TestCase`

crynobone:
We don't typically change this configuration. It shouldn't be in the stubs file.


driesvints:
@crynobone tests fail here

PR #1099, opened 2024-04-16
Update GitHub Actions workflow

crynobone:


driesvints:
@crynobone tests fail here

PR #1101, opened 2024-04-25
update DuskTestCase.stub for define github or local

tszyuloveyou:
For https://github.com/laravel/docs/pull/9601
ref: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -87,11 +87,8 @@ jobs:
           composer update "laravel/dusk" --prefer-dist --no-interaction --no-progress -W
           php artisan dusk:install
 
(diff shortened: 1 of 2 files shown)


driesvints:
Thanks. I'm not sure why the stub changes were necessary. I removed these.

driesvints:
Actually, I'm gonna revert this one because this step actually tested the command itself which we aren't doing anymore after this PR.

crynobone:
I assume this changes to make way for alternative ChromeDriver such as Firefox and Safari but we should see an actual use case first and maybe look at how Livewire project does it with Testbench Dusk

PR #1108, opened 2024-07-23
fix: DuskTestCase driver function fails to read the DUSK_DRIVER_URL environment variable

### LINKED ISSUE #1107: The driver fails to read the environment properly
### Dusk Version

11.13.0

### Laravel Version

11.13.0

### PHP Version

8.2

### PHPUnit Version

10.5.17

### Database Driver & Version

_No response_

### Description

Dusk determine the selenium server url by searching for a DUSK_DRIVER_URL variable in $_ENV. This works properly with Sail when issuing a `sail dusk` (one can even specify a custom driver url with a `DUSK_DRIVER_URL=http://my-custom-url:4444`).

However this fails in several environment : 
- tested personally in a local environment with a "default" php-fpm setup
- in Github Actions

After a little bit of research, the culprit is the use of `$_ENV` (see this [stackoverflow question](https://stackoverflow.com/questions/3780866/why-is-my-env-empty)).

In both of these configurations the env variables are defined `$_SERVER`, not `$_ENV`. 

Can this be considered a "bug" or is it the expected behavior ? I can see a couple solutions : 
- use `getenv('DUSK_DRIVER_URL')` instead of `$_env['DUSK_DRIVER_URL']`
- use a combination of $_ENV and $_SERVER to search for DUSK_DRIVER_URL

I'm not sure whether it would open security issues tho, with regard to loading an environment variable from a broader context ... Although I don't see how loading an environment variable from another context for Dusk test execution, even in production could lead to a security breach (RCE and so on).

I modified my DuskTestCase for my project to fix my issue, however if you see value in making this change to the base repo I'd be glad to submit a PR 

### Steps To Reproduce

1. setup a default laravel project 
2. install Dusk
3. modify `DuskTestCase`, line 40, to specify an invalid URL ([here](https://github.com/laravel/dusk/blob/8.x/stubs/DuskTestCase.stub#L42))
4. run `DUSK_DRIVER_URL=http://localhost:9515 php artisan dusk`, Dusk should use the invalid URL

robinstraub:
resolves #1107

### DIFF
--- a/stubs/DuskTestCase.stub
+++ b/stubs/DuskTestCase.stub
@@ -39,7 +39,7 @@ abstract class DuskTestCase extends BaseTestCase
         })->all());
 
         return RemoteWebDriver::create(
-            $_ENV['DUSK_DRIVER_URL'] ?? 'http://localhost:9515',
+            env('DUSK_DRIVER_URL') ?? 'http://localhost:9515',
             DesiredCapabilities::chrome()->setCapability(
                 ChromeOptions::CAPABILITY, $options
             )
--- a/stubs/DuskTestCase.stub
+++ b/stubs/DuskTestCase.stub
@@ -39,7 +39,7 @@ abstract class DuskTestCase extends BaseTestCase
         })->all());
 
         return RemoteWebDriver::create(
-            env('DUSK_DRIVER_URL') ?? 'http://localhost:9515',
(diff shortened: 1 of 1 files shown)


taylorotwell:
i would probably still keep $_ENV but just fall back to env() then fallback to the last case.

Issue #1113, opened 2024-08-05
Dusk @-selectors cannot be combined with other CSS selectors for the same element


pindab0ter:
### Dusk Version

8.2.2

### Laravel Version

11.11.1

### PHP Version

8.2.22

### PHPUnit Version

n/a

### Database Driver & Version

n/a

### Description

When using Dusk `@`-selectors, they cannot be combined with for other CSS selectors that are applicable to the same element.

For example: I have an HTML element with `class="foo" dusk="bar"`. I want to match an element that has `@bar` as a Dusk selector and also `.foo` as a class. Also checking for a CSS class is useful for example when using Dusk Components to make sure the Dusk selector is applied to the correct element:

```php
class TestComponent extends Component
{
    public function __construct(
        private readonly string $selector
    ) {
    }

    public function selector(): string
    {
        return "{$this->selector}.bar";
    }

   …
}
```

After parsing, the selector (here `@foo`) becomes `[dusk="foo.test"]` instead of `[dusk="foo"].test`. This is due to [this regex](https://github.com/laravel/dusk/blob/971907bdd6e50e32a0855ab3d21576ebef0f957f/src/ElementResolver.php#L413) only looking for spaces, not other CSS selector symbols like `.` and `#`.

A possible solution would be to change the regex to also look for CSS selector symbols in addition to spaces. This would break any existing Dusk selectors that contain periods (e.g. `dusk="@foo.bar"`) or other CSS selectors. As of now that is valid, but it is my opinion that they should not be. In my opinion people should use commonly used separators like dashes or underscores in Dusk selectors, just like you would in `id` and `class` names.

Right now it is not possible to distinguish whether `.bar` in `@foo.bar` refers to a class named `bar` or is a part of the Dusk selector itself (e.g.: `<div class="bar" dusk="foo.bar">`).

Trying to work around this limitation by putting the Dusk selector after the other CSS selectors doesn't work either, as the parser looks for `@` being the first character in the selector in the line above the previously linked line.

While I realise it might not be intended to use Dusk selectors in combination with other CSS selectors, I do believe this adds value and should be supported. To that end I suggest deprecating using reserved CSS symbols in Dusk selectors and updating the format regex accordingly.

### Steps To Reproduce

1. Apply a class name and a Dusk selector to an element (e.g. `class="foo" dusk="bar"`)
2. Write a selector that includes both (e.g. `assertVisible('@bar.foo')` or `assertVisible('.foo@bar')`)

[...]

PR #1114, opened 2024-08-08
Require Dusk selectors to adhere to CSS naming rules

pindab0ter:
# Description

This PR fixes issue #1113 by refining the Dusk selector parsing to ensure compatibility with valid CSS class names. Currently, the documentation, PHPDoc, and unit tests do not clearly define which characters are valid for Dusk selectors, leading to potential conflicts when Dusk selectors are combined with other CSS selectors on the same element.

# Motivation
The existing documentation and unit tests only uses dashes in Dusk selectors. However, Dusk selectors cannot be used alongside other CSS selectors if they target the same element, which requires them not to be separated by a space. By aligning the Dusk selector parsing with the standards for valid CSS names (as outlined in [this Stack Overflow answer](https://stackoverflow.com/questions/448981/which-characters-are-valid-in-css-class-names-selectors#449000)), we add support for Dusk selectors to target the same element as other CSS selectors.

# Changes

* Updated the Dusk selector parsing regex to only allow valid CSS names.
* Modified the existing condition to use Str::contains instead of Str::startsWith to support Dusk selectors to exist after another CSS selector not separated by a space.
* Added unit tests to ensure the new regex handles combined selectors correctly.

### DIFF
--- a/src/ElementResolver.php
+++ b/src/ElementResolver.php
@@ -409,8 +409,8 @@ class ElementResolver
             array_keys($sortedElements), array_values($sortedElements), $originalSelector = $selector
         );
 
-        if (Str::startsWith($selector, '@') && $selector === $originalSelector) {
-            $selector = preg_replace('/@(\S+)/', '['.Dusk::$selectorHtmlAttribute.'="$1"]', $selector);
+        if (Str::contains($selector, '@') && $selector === $originalSelector) {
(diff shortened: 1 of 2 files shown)


taylorotwell:

pindab0ter:
Can I get a response that goes into the core issue of the problem this is trying to solve?

This feels like a canned response, as this PR does only changes one line of code and adds two unit test cases.

u01jmg3:
@pindab0ter: might be worth flagging to @driesvints ☮️

PR #1115, opened 2024-08-11
[8.x] Extract chromedriver regardless of position in archive

aleahy:
The aim of this PR is to do a small refactor to prevent dusk breaking if Chrome annoyingly change the order of, or add files in their driver zip archive.

It looks through the archive and only extracts files beginning with `chromedriver`. This should handle all windows, linux and mac chromedriver files.

### DIFF
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -94,6 +94,7 @@ class ChromeDriverCommand extends Command
         foreach (OperatingSystem::all() as $os) {
             if ($all || ($os === $currentOS)) {
                 $archive = $this->download($version, $os);
+
                 $binary = $this->extract($archive);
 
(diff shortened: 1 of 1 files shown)


crynobone (approved):

src/Console/ChromeDriverCommand.php
```diff
@@ -217,26 +216,34 @@ protected function download($version, $os)
      * @param  string  $archive
```
u01jmg3, line 216: ☝️ docblock needs to updated

aleahy: Thanks! Updated the extract docblock.

u01jmg3: The other thing I noticed is elsewhere in Dusk they state `@throws \Exception`.

PR #1116, opened 2024-08-11
[7.x] Fix for change in chromedriver archive structure

aleahy:
This PR fixes the bug that exists in version 7.x due to the change in the archive structure for the chrome-driver.

### DIFF
--- a/src/Console/ChromeDriverCommand.php
+++ b/src/Console/ChromeDriverCommand.php
@@ -235,7 +235,6 @@ class ChromeDriverCommand extends Command
             }
         }
 
-
         $zip->close();
 
         unlink($archive);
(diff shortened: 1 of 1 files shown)


crynobone (approved):

src/Console/ChromeDriverCommand.php
```diff
@@ -215,20 +215,34 @@ protected function download($version, $os)
      * @param  string  $archive
```
u01jmg3, line 214: ☝️ docblock needs to updated

driesvints:
@aleahy v7 isn't maintained any longer. Please upgrade to v8

u01jmg3:
@driesvints: should GitHub Actions be set to autoclose PRs raised against branches no longer receiving support?

driesvints:
@u01jmg3 I've thought of developing such a flow but haven't found the time yet.

driesvints:
Can maybe become part of https://github.com/laravel/.github/blob/main/.github/workflows/pull-requests.yml as a new `auto-close` flow that closes PR's against branches that aren't the default branch. Also, maintainers should still always be able to send PR's to any branch.

PR #1143, opened 2024-10-11
check for directory existence when storing console log or storing source, if it does not exist create it

ssanko:
This PR fixes missing destination directory for storing console log or source code.

Currently when saving console log or source code to the path defined in `\Laravel\Dusk\Browser::$storeConsoleLogAt` or `\Laravel\Dusk\Browser::$storeSourceAt`, the code does not check for the existence of the directory.

Because of missing check for destination directory existence, current code throws `ErrorException` when trying to save console log or source code:
```
file_put_contents(/Sites/laravel-dusk/storage/app/customPath/pageSources/source.txt): Failed to open stream: No such file or directory
```

This PR adds code to check for the exitence of the destination directory and if it does not exist, then tries to create it.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -484,8 +484,16 @@ class Browser
             $console = $this->driver->manage()->getLog('browser');
 
             if (! empty($console)) {
+                $filePath = sprintf('%s/%s.log', rtrim(static::$storeConsoleLogAt, '/'), $name);
+
+                $directoryPath = dirname($filePath);
+
+                if (! is_dir($directoryPath)) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Any reason why you use 0777 permissions instead of 0755?

ssanko:
I wanted to stick to the logic used in method `\Laravel\Dusk\Browser::screenshotElement` which uses the 0777 permission.

So if the 0755 is more reasonable, then I'll change the permissions also for `\Laravel\Dusk\Browser::screenshotElement` method.

Issue #1150, opened 2024-12-26
assertSee() / assertSeeIn() doesn't support selectors to multiple elements 


flap152:
### Dusk Version

8

### Laravel Version

11.34

### PHP Version

8.3

### PHPUnit Version

11

### Description

The current implementation of assertSee() / assertSeeIn() / assertDontSee / assertDontSeeIn() support selectors to multiple elements but assume only one element is selected while there are valid/legitimate selector that returns multiple elements. 
For example, the content of the first column in a dataTable, ignoring content in other columns: `#crudTable > tbody > tr > td:nth-child(1) > span`. 

I find only the first element is considered and assert on it if it exists.
The asserts do not provide feedback when multiple elements are returned and will be right or wrong depending if the target text is present in the first or in other elements.

After making a macro for additional asserts methods, I found the existing code could be extended to manage these cases relatively easily.

The contribution guide says a "draft" PR will not be reviewed, but I seek opinions on choices available in implementing the changes, I will link to my fork of dusk, or will a Draft PR be useful here since it is referenced in an issue?

https://github.com/flap152/dusk/pull/1

### Steps To Reproduce

Given this html content:
```
<!DOCTYPE html>
<html>
    <head>
        <title>Assert See In</title>
    </head>
    <body>
        <div id="app">
            <div>First</div>
            <div>Second</div>
            <div>Third</div>
            <div>Fourth</div>
            <h1>Hello World</h1>
        </div>
    </body>
</html>
```
the following asserts would fail with either a false negative or "element not found"

```
    public function test_it_handles_assert_see_in_with_multiple_selection()
    {
        $this->browse(function (Browser $browser) {
            $browser->visit('/tests/assert-see-in');
             $browser->assertSeeIn('#app > div:nth-child(3)', 'Third');
             $browser->assertDontSeeIn('#app > div:nth-child(3)', 'Second')
            ;
        });
    }
```

flap152:
I submitted PR and made it ready for review. Hope I'm not too far off. :cold_sweat: :crossed_fingers:

flap152:
Hi @crynobone, the PR is closed/delayed for now and that's ok, but I'd really appreciate feedback on the issue (behavior with multiple elements), but also the submitted changes, comments and questions I made regarding the implementation and testing.
I think this would help make valuable contributions to dusk/laravel in the future.

PR #1151, opened 2024-12-28
Support multiple elements from selector in assertSee and variants (Dont and/or In)

flap152:
The current implementation of assertSee() / assertSeeIn() / assertDontSee / assertDontSeeIn() support selectors to multiple elements but assume only one element is selected while there are valid/legitimate selector that returns multiple elements.
For example, the content of the first column in a dataTable, ignoring content in other columns: #crudTable > tbody > tr > td:nth-child(1) > span.

I find only the first element is considered and assert on it if it exists.
The asserts do not provide feedback when multiple elements are returned and will be right or wrong depending if the target text is present in the first or in other elements.

After making a macro for additional asserts methods, I found I needed the other (In and/or Dont) variants and the existing code could be extended to manage these cases relatively easily.

The changes transparently support multiple elements

Questions/choices:

- Current unit tests use mocks with implementation knowledge and needed changes.

  - I found using cardinality specifiers (once(), twice(), atLeast(), etc.) made tests more precise but may become brittle?
  - I duplicated the test for whenAvailable(), then split it for each cases: element found / not-found. This helped me understand the mocked method usage and may be preferable, or not 😉 
- Failed assertion text can be updated to reflect possible multiples, but tests have hard-coded expectations and existing Dusk user tests may also "expect" them as-is.

- I wonder about requirement/behavior to fail if no element selected vs searched text being found/absent. It seems it is partially just handling exception from the driver. Maybe I need clarification on that.

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -189,6 +189,7 @@ trait MakesAssertions
 //          "Did not see expected text [{$text}] within any of the elements [{$selector}]."
             "Did not see expected text [{$text}] within element [{$fullSelector}]." //hard coded in test
         );
+
         return $this;
     }
 
@@ -208,12 +209,13 @@ trait MakesAssertions
 
         PHPUnit::assertNotEmpty($elements);
 
-        foreach ($elements as $element){
+        foreach ($elements as $element) {
             PHPUnit::assertFalse(
                 Str::contains($element->getText(), $text, $ignoreCase),
                 "Saw unexpected text [{$text}] within element [{$fullSelector}]."
             );
         }
+
         return $this;
     }
 
@@ -231,7 +233,7 @@ trait MakesAssertions
 
         PHPUnit::assertNotEmpty($elements);
 
-        foreach ($elements as $element){
+        foreach ($elements as $element) {
             PHPUnit::assertTrue(
                 $element->getText() !== '',
                 "Saw unexpected text [''] within element [{$fullSelector}]."
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -189,6 +189,7 @@ trait MakesAssertions
 //          "Did not see expected text [{$text}] within any of the elements [{$selector}]."
             "Did not see expected text [{$text}] within element [{$fullSelector}]." //hard coded in test
         );
+        return $this;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


flap152:
Mentioned in #1150 (assertSee() / assertSeeIn() doesn't support selectors to multiple elements)

tests/Unit/WaitsForElementsTest.php
```diff
@@ -391,16 +442,19 @@ public function test_wait_for_an_element_to_be_disabled()
     public function test_wait_for_text_in()
     {
         $element = m::mock(RemoteWebElement::class);
-        $element->shouldReceive('getText')->andReturn('Discount: 20%');
+        $element->shouldReceive('getText')->once()->andReturn('Discount: 20%');
 
         $driver = m::mock(WebDriver::class);
+//        $driver->shouldReceive('wait')->with(5, 100)->andReturnUsing(function ($seconds, $interval) use ($driver) {
```
flap152, line 448: I found wait() was sometimes caled with 5 and not 2, and had to use this commented out version. Could not understand why.
Just realized it's probably because of  `test_default_wait_time()` test higher up, setting static value that persists within the class (`Browser::$waitSeconds = 2;`), but sometimes not executed - when filtering only this test for example.

It should probably be reset, either in that test (`try ... finally{Browser::$waitSeconds = 5;}`) or in the class tearDown() method

src/Concerns/MakesAssertions.php
```diff
@@ -176,12 +176,18 @@ public function assertDontSee($text, $ignoreCase = false)
     public function assertSeeIn($selector, $text, $ignoreCase = false)
     {
         $fullSelector = $this->resolver->format($selector);
+        $elements = $this->elements($selector) ?? [];
+//        PHPUnit::assertNotEmpty($elements);
```
flap152, line 180: AssertNotEmpty is not strictly required because the assert below will fail, but the "exception" or reason may be different and this may not be desired

tests/Unit/MakesAssertionsTest.php
```diff
@@ -1346,7 +1346,33 @@ public function test_assert_dont_see()
         }
     }
 
-    public function test_assert_see_in()
+//    public function test_assert_see_in()
```
flap152, line 1349: Duplicated, commented out, and renamed below when I thought more than one test would be needed. This should go away and keep the original test name, I guess.

tests/Unit/WaitsForElementsTest.php
```diff
@@ -53,6 +55,55 @@ public function test_when_available()
         }
     }
 
+    public function test_when_available_missing()
```
flap152, line 58: These 2 tests reproduce the 2 use cases in `test_when_available()`, when the item is found and when it's not found.
Although these new tests have some duplication in the mock setup, they are easier to understand, which I think is better, considering the inherent complexity of testing a method that uses a callback

taylorotwell:

PR #1163, opened 2025-02-11
feat: wait for query string

Barbapapazes:
Hello 👋,

There is an assert called assertQueryStringHas but nothing to wait for a specific query string.

Actually, I have an input that will change the URL query string before showing the new data and having a way to wait for a change in the query string could be very useful.


taylorotwell:

taylorotwell:
It feels like we should re-use or be consistent with how we pull / inspect the query string across these methods.

PR #1166, opened 2025-02-19
[9.x] replace jQuery with vanilla JS

pataar:
This removes the need of jQuery all together and replaces it with Vanilla JS calls.

It has a couple of advantages:
- No update management of external libraries needed anymore
- Should be more performant because jQuery isn't loaded anymore
- Ensures that user-loaded jQuery libraries do not conflict (when the user loads an older jQuery version for example)


crynobone:
This is a breaking change, please retarget to `master` branch

pataar:
Done!

PR #1167, opened 2025-02-20
Disable smooth scrolling in TestCase

ziadoz:
My colleague @MizouZie tracked down that this behaviour can be the source of flaky test failures, particularly in CI. We turned this flag on a few months ago, and it's reduced our intermittent CI test failures significantly.

The flag is documented here: https://peter.sh/experiments/chromium-command-line-switches/#disable-smooth-scrolling

The type of failures we've found it helps with are ones like this, when locating the element involves scrolling:

```
NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":"body #foo-bar"}
```


taylorotwell:
Awesome - thanks

Issue #1006, opened 2022-11-10
Custom chromedriver install path


underdpt:
Hello,

While using dusk for browser automation with [Laravel-Zero](https://github.com/laravel-zero/laravel-zero) we found it's not possible to run the chromedriver when the laravel-zero app is built into a phar file, as all vendor files are copied into the phar file, including the chromedriver binaries.

This also generates other issues, like the final user being unable to install the chromedriver if his chrome version changes.

I would like to send a PR to allow specifying the chromedriver install path, being that path used also when creating the process. I'm not sure what would be the best way to specify the install path. Would a `config/dusk.php` configuration file be fine for you? I think a configuration file would open other possibilities (like configuring the sreenshots and logs path, setting custom options for the driver, etc.).

Related: https://github.com/laravel-zero/laravel-zero/issues/193#issuecomment-1310197874

driesvints:
Feel free to attempt a PR and we'll see how it goes 👍

PR #1171, opened 2025-04-10
Allows Chrome driver's path to be configured

Max13:
Allows to customize Chrome driver path using a `dusk.php` configuration file. Can be useful when building a phar standalone app (ie. Laravel-zero)

Implements #1006


taylorotwell:

PR #1179, opened 2025-10-31
[8.x] do not pause initially while waiting for an element

pataar:
Currently the `waitFor` calls are pausing for `$interval` before checking that an element is visible. In a lot of cases this pause is obsolete because the element could be visible initially. 

Potentially this could speed up many tests where a lot of `waitFor` calls are done while elements are visible initially.


taylorotwell:

pataar:
@taylorotwell Will Dusk be going to an EOL state in the future?

PR #1197, opened 2026-04-14
[8.x] Me again, with `assertAcceptingDialog/Dismissing()`

weshooper:
I know we must wait for dialogs, but don't wanna read all about it in our tests.

Maybe other Dusk aficionados feel the same? 🤓

No tests as super boring proxies, that I pinky promise to add to docs...

### DIFF
--- a/src/Concerns/MakesAssertions.php
+++ b/src/Concerns/MakesAssertions.php
@@ -987,6 +987,36 @@ JS;
         );
 
         return $this;
+    }    
+
+    /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
The naming makes me think I'm asserting some previous action, but the method actually accepts the dialog actively.

weshooper:
I've renamed to be more active.

taylorotwell:
Hmm, I dunno, I may skip this one.

